PR #40, opened 2016-08-20
Internationalise thing status messages

cdjackson:


cdjackson:
@maggu2810 any thoughts on this would be welcome before I merge

OSGI-INF/ZWaveBindingConstants.xml
```diff
@@ -0,0 +1,13 @@
...
+    Copyright (c) 2014-2016 by the respective copyright holders.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+-->
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" name="org.openhab.binding.zwave.ZWaveBindingConstants">
+   <implementation class="org.openhab.binding.zwave.ZWaveBindingConstants"/>
+
+   <reference bind="setI18nProvider" cardinality="1..1" interface="org.eclipse.smarthome.core.i18n.I18nProvider" name="I18nProvider" policy="static" unbind="unsetI18nProvider"/>
```
maggu2810: If you are using 0..1 there is no need to restart this service if the I18nProvider service disappears and appears again (e.g. by a bundle restart etc.). WRT the current bundle location of that service it doesn't really make a difference, but as you can go ahead without the i18n provider service, perhaps you want to change that.

src/main/java/org/openhab/binding/zwave/ZWaveBindingConstants.java
```diff
@@ -89,14 +92,58 @@
...
+
+    protected void unsetI18nProvider(I18nProvider i18nProvider) {
+        ZWaveBindingConstants.i18nProvider = null;
+    }
+
+    public static String getI18nConstant(I18nConstant constant) {
+        if (i18nProvider == null) {
+            return MessageFormat.format(constant.key, (Object[]) null);
+        }
+        return i18nProvider.getText(ZWaveActivator.getContext().getBundle(), constant.key, constant.defaultText, null,
```
maggu2810: If you want to prevent a NPE in every case (I think you will perhaps never run into one), you have to change the code a little bit, because the unset method could be called async between the check and this line.

maggu2810:
IMHO the code is okay. I personally try never to use static methods that access non-static members. Another option would be that your thing handler factory depends on the i18n provider and the code you add to the binding constant will be in another utility object. This object could be injected to the thing handler on creation.
But as the i18n provider is living in some core bundles, it does not matter at all. Its just about personal feelings: :wink: 

So, no any further thoughts.

cdjackson:
Thanks - updated.

I think I’ll leave the static methods since you don’t object too much ;)

src/main/java/org/openhab/binding/zwave/ZWaveBindingConstants.java
```diff
@@ -89,14 +92,60 @@
...
+    protected void unsetI18nProvider(I18nProvider i18nProvider) {
+        ZWaveBindingConstants.i18nProvider = null;
+    }
+
+    public static String getI18nConstant(I18nConstant constant) {
+        I18nProvider i18nProviderLocal = i18nProvider;
+        if (i18nProviderLocal == null) {
+            return MessageFormat.format(constant.key, (Object[]) null);
+        }
+        return i18nProvider.getText(ZWaveActivator.getContext().getBundle(), constant.key, constant.defaultText, null,
```
maggu2810: I assume you want to use the local copied reference `18nProviderLocal` here

src/main/java/org/openhab/binding/zwave/ZWaveBindingConstants.java
```diff
@@ -89,14 +92,60 @@
...
+                (Object[]) null);
+    }
+
+    public static String getI18nConstant(I18nConstant constant, Object... arguments) {
+        I18nProvider i18nProviderLocal = i18nProvider;
+        if (i18nProviderLocal == null) {
+            return MessageFormat.format(constant.key, arguments);
+        }
+        return i18nProvider.getText(ZWaveActivator.getContext().getBundle(), constant.key, constant.defaultText, null,
```
maggu2810: I assume you want to use the local copied reference 18nProviderLocal here

cdjackson: Ah - damn - well spotted :)

PR #45, opened 2016-08-21
Update discovery to ExtendedDiscoveryService

cdjackson:
Improves management of device discovery - devices should now not be overwritten if we have less information than is already in the inbox.


cdjackson:
Note that this is a breaking change in that it requires an update to the latest OH2 runtime.

PR #119, opened 2016-09-21
Set battery node to DEAD if no response within 2x wakeup period

### LINKED ISSUE #55: Feature request: Periodically check for device ONLINE/OFFLINE status based on wakeup interval settings
Greetings,

as discussed [on the forum](https://community.openhab.org/t/z-wave-battery-devices-issues-in-general/10022/52), it would be desirable for the ZWave binding to update the Thing status for ZWave devices by using the wakeup interval and last wakeup information already available.
This is especially true in battery-powered devices, as they currently can physically disappear from the network (e.g. battery removed or completely discharged) while keeping their "ONLINE" status indefinitely.
I guess this enhancement could be implemented by checking periodically, for every ZWave Thing currently ONLINE, the condition
`NOW - (zwave_wakeup_time + wakeup_interval) > SET_OFFLINE_TIME_THRESHOLD`
where
 - `NOW` would be the current time
 - `zwave_wakeup_time` is the Thing property reporting the last time a wakeup signal from the device was observed
 - `wakeup_interval` is how often the Thing performs a wakeup (in the Thing configuration settings)
 - `SET_OFFLINE_TIME_THRESHOLD` would be a tolerance threshold (ideally close to 0) after which the Thing is to be considered definitely offline.

On a side note, we observed that mains-powered devices are not set OFFLINE when removed from the power source unless we try to update their status: is this expected behavior?

cdjackson:
fixes #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ ZWave is a wireless home automation protocol with reliable two way communication
 
 A wide range of devices are supported from lights, switches and sensors to smoke alarms, window coverings and keyfobs. Z-Wave certification guarantees that certified devices will be compatible with each other and the network.
 
-The binding uses a standard Z-Wave stick to communicate with the Z-Wave devices. There are many sticks available, and they all support the same interface so the binding does not distinguish between them.
+The binding uses a standard Z-Wave serial stick to communicate with the Z-Wave devices. There are many sticks available, and they all support the same interface so the binding does not distinguish between them.
 
 
 ## Supported Things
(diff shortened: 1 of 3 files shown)


xsnrg:
Chris, this is merged and not a big deal, but the ZCOMBO devices don't have a settable wake period.  It is a strange read-only thing.  I imagine they do this for maximum battery.  I don't have any documentation that says what the wake period is either, so this device is now marked as dead sometimes.  My work around has been to set the wake interval long enough that it wakes within 2 periods.  I am testing this to see if it marks it as dead again, but as you could imagine, there is a long testing cycle :-)  

As long as messages coming from the node are not ignored because of the dead status, functionally it should be fine, but confusing.  I will look through the log to see if I can see the interval between wakes.  If I can figure this out, is it possible to set a default wake interval for the device in the database?

cdjackson:
Hi Jim,
Ok - maybe I need a way to disable this, although if you can manually set the wakeup to a long number, even if it’s ignored by the device, then this should do the same thing. The binding won’t ignore any messages, and any message received should make the device alive again…

If it’s a problem, and you can’t keep the device alive then open a new issue and we can discuss a workaround.

Cheers
Chris

xsnrg:
Hi Chris, 

This seems to be working.  I set it to a period long enough that it does not show the device as dead any longer.  The battery status is updated, so I think we are good.

Thanks
Jim

PR #138, opened 2016-09-27
Update philio_pan08_0_0.xml

vectah0:
channel added - blinds_control


cdjackson:
As said by @xsnrg changes should be made to the online database otherwise they will be overwritten when the database gets updated.

PR #148, opened 2016-10-02
Create philio_pst02-1c_0_0.xml

schmidmuc:
Should be more or less the same device as PST02C


cdjackson:
This device is already in the database and any changes need to be made to the database itself or they will be lost when the database is exported

cdjackson:
FTR this appears to be the same device based on IDs. http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/170

PR #149, opened 2016-10-02
Create philio_phiepat02_1_8.xml

schmidmuc:


cdjackson:
Please can you add these to the online database -:

http://www.cd-jackson.com/index.php/zwave/zwave-device-database

-: If I merge this it will be deleted next time the database gets updated since the database is exported automatically.

schmidmuc:
Hi chjackson, I tried to add it to the online database, but I don't find or don't have the possibility to add a device. The "add" icon is not shown...

cdjackson:
I need to give you access (it should say to email me in the instructions). I’ll see if I can find your username and update your access now.

cdjackson:
Done - you should be able to edit now…   Let me know if you have any problems.

schmidmuc:
Thanks... I tried but got the following error:

1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ',,705)' at line 3 SQL=INSERT INTO `joo_zwave_endpoints` (`device`,`number`,`label`,`basicclass`,`genericclass`,`specificclass`,`created_by`) VALUES (390,0,'Endpoint 0',,,,705)

cdjackson:
What XML are you trying to load - is this XML file created by OH (i.e. from the userdata/zwave folder)?

schmidmuc:
No, I don't have a XML in the userdata folder for this device...
I copied the XML file I added here in Github

cdjackson:
The database can’t read these files - it creates them. It needs the file from openhab to read the device configuration first. If the file isn’t being created, we need to work out why - I’d suggest to grab a debug log.

schmidmuc:
How can I update the zwave database without getting a new openhab2 or zwave binding update?
Do I have to exchange a certain file?

cdjackson:
You need to update the binding as the database is compiled into the binding. To do this, use either PaperUI or HABmin, select Extensions, click on ZWave Binding, then uninstall, and install again - this will install the most recent version.

Note that this assumes you are using the online snapshot version. If you are using an offline version, then you need to update the whole runtime.

PR #171, opened 2016-10-12
Database resynchronisation

cdjackson:


xsnrg:
I had a few devices that were changed in the DB but even though they were approved, they never made it into the repo.  Thanks for bringing it back to parity.  If I see it happen again, I will open a ticket.

cdjackson:
Thanks Jim.

PR #247, opened 2016-12-01
Transaction refactoring + SECURITY

cdjackson:
This is an initial PR to implement improved transaction handling. It is a large rewrite of the lower layers of the binding and uses futures to allow async transaction handling. It also has major changes to command class handling.

There's been a focus on trying to remove, or consolidate conditional code. For example all queuing is handled in a single class rather than queueing battery device messages in another place. This all helps ensure we only handle things in a single place which is especially useful relating to security encap...

So... This also adds the S0 SECURITY class.

SECURITY is still very much WIP - it is a complete rewrite from OH1 and a major simplification. It seems to work well, but hasn't been tested for robustness yet. It's not been tested on battery devices either (or my door lock) - testing has focused on using an Aeon multisensor connected to USB. Security does handle restarts quite well it seems - even if the XML is lost.

A number of other changes are addressed here - it's likely to still be buggy due to the large number of changes. It's also full of debug messages which will be removed before merge - be warned - logs might fill up if you use this :)

This code changes the command class names to use the standard ZWave names - as such, the database doesn't work (except maybe the Fibaro RGBW that I've been using for testing!). The database can easily be changed to export with the new names.

Device classes probably should also be renamed before this is merged - do the whole lot once!!!


kaikreuzer:
Your branch looks quite messy with 115 commits in this PR - I hope you will use a "squash" when merging ;-)

cdjackson:
Yes - it’s very much a development branch ;). As mentioned, it’s a major refactoring of the binding….

kaikreuzer:
>  As mentioned, it’s a major refactoring of the binding….

Yes, I saw that - I think many people will be happy about the security class 👍

mhilbush:
@cdjackson  I have some to test this.  Any issue with me grabbing the latest jar from the PR build on CB?

cdjackson:
I think the version on git won't work at the moment as I don't think it includes a compatible database. I'll push an ipdate over the weekend to fix that. 

There might still be some issues to work through with battery devices as well but let's see. 

I'd certainly appreciate the testing so thanks for the offer. 

Chris

mhilbush:
Ok, just let me know when you think it's ready.

pizerg:
Hi,

I'm going to start the testing of this PR tomorrow, should we open separate issues for feedback or we post them here?

Adrián

cdjackson:
I've created a ```development``` branch and will merge this into that branch. I'll close this and reopen a new PR against the new branch shortly.

PR #272, opened 2016-12-21
put detection of O/D endpoint swap back in.

jspuij:
- Added logging for received encapsulated frames where the destination
endpoint on the controller is > 1, as the controller in the binding has
no endpoints.
- Added a warning for received encapsulated frames where in addition to
the condition above, the originating endpoint is either 0 (root) or 1
(first endpoint on originating node), as this is a firmware bug with
high probability, where originating and destination are swapped.
- Swap only if conditions above are true and option
useDestEndpointAsSource is set. (And don’t log, as this happens at a
lot of frames and since the option is set, it must be intentional).

- Get command class from node root if originating endpoint = 0. Log
this as encapsulating a message that originates from the node root (and
probably is destined for the root or first endpoint of the controller)
is useless.

Signed-off-by: Jan-Willem Spuij jw@spuij.nl (github: jspuij)

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveMultiInstanceCommandClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveMultiInstanceCommandClass.java
@@ -453,10 +453,34 @@ public class ZWaveMultiInstanceCommandClass extends ZWaveCommandClass {
         int originatingEndpointId = serialMessage.getMessagePayloadByte(offset);
         int destinationEndpointId = serialMessage.getMessagePayloadByte(offset + 1);
 
-        if (useDestEndpointAsSource) {
-            // Not a full swap. Do not use destinationEndpointId after this line
-            // and leave scope intact.
-            originatingEndpointId = destinationEndpointId;
+        // if a specific endpoint on the controller is addressed.
+        if (destinationEndpointId > 1) {
+            // swap specified in node options and condition satisfied:
+            if (useDestEndpointAsSource && originatingEndpointId <= 1) {
+                // swap originating and destination.
+                int temp = originatingEndpointId;
+                originatingEndpointId = destinationEndpointId;
+                destinationEndpointId = temp;
+            } else {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveMultiInstanceCommandClass.java
```diff
@@ -453,10 +453,34 @@ private void handleMultiChannelEncapResponse(SerialMessage serialMessage, int of
...
+            if (useDestEndpointAsSource && originatingEndpointId <= 1) {
+                // swap originating and destination.
+                int temp = originatingEndpointId;
+                originatingEndpointId = destinationEndpointId;
+                destinationEndpointId = temp;
+            } else {
+                // received an encapsulation for an endpoint on the controller.
+                logger.info(
+                        "NODE {}: Received a multi instance encapsulation with a destination endpoint = {}. "
+                                + "The controller in our binding does not have endpoints.",
```
cdjackson: Can we reduce the log message to one line - the controller might have endpoints one day so we shouldn't make this statement.

src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveMultiInstanceCommandClass.java
```diff
@@ -470,21 +494,34 @@ private void handleMultiChannelEncapResponse(SerialMessage serialMessage, int of
...
-        if (endpoint == null) {
-            logger.error("NODE {}: Endpoint {} not found. Cannot set command classes.", getNode().getNodeId(),
-                    originatingEndpointId);
-            return;
-        }
+        if (originatingEndpointId == 0) {
+            // originating endpoint == 0, get the command class from the root node.
+            // encapsulation was useless.
+            logger.warn(String.format(
+                    "NODE %d: Originating Command Class %s (0x%02x) was on the root node."
```
cdjackson: Can we reduce the log message to the first line here.

PR #293, opened 2017-01-02
Update notification converter to filter events

cdjackson:


mhilbush:
Would the approach you took here also apply to the ZW100 issue that we discussed here?

https://community.openhab.org/t/proper-openhab-2-configuration-for-aeon-multisensor-6-zw100/18933/8

cdjackson:
Probably, yes - I had intended to also resolve the ZW100 last night, but I had problems getting it to initialise with the current branch. This should allow distinction of the different alarms though, so should solve these issues...

PR #297, opened 2017-01-03
Zipato MINI KEYPAD RFiD/ZWAVE: change alarm_access to notification_ac…

DocFraggle:
Hi Chris, as discussed here: https://community.openhab.org/t/wtrfid-mini-keypad-rfid-z-wave/19014/27 
I changed the channel and tested it locally, the channel is updated properly when pressing the buttons.

Regards, Christian

### DIFF
--- a/ESH-INF/thing/wintop_wtrfid_0_0.xml
+++ b/ESH-INF/thing/wintop_wtrfid_0_0.xml
@@ -23,10 +23,10 @@
           <property name="binding:*:OnOffType">ALARM;type=BURGLAR</property>
         </properties>
       </channel>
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks - I'm pleased it worked :)

While this looks ok, it really should be added directly into the database. If we merge this, it will be removed again when the database is next exported as the database version will overwrite your XML file changes.

This probably isn't what you want ;)

Editing the XML manually is fine for testing, but for merging into the database, really the change needs to go into the online system - do you mind doing that? If not I can do it.

DocFraggle:
Ah, sure, no problem!

PR #352, opened 2017-02-08
Added missing meter_reset channels

TheNetStriker:
I’ve added some missing meter_reset channels. I already tested this and
it works perfectly. Please add this to the main repository.

Signed-off-by: David Masshardt <david@masshardt.ch> (github:
TheNetStriker)

### DIFF
--- a/ESH-INF/thing/aeon_zw075_0_0.xml
+++ b/ESH-INF/thing/aeon_zw075_0_0.xml
@@ -41,6 +41,12 @@
           <property name="binding:*:DecimalType">METER;type=E_KWh</property>
         </properties>
       </channel>
+      <channel id="meter_reset" typeId="meter_reset">
+        <label>Reset the total power consumption</label>
+        <properties>
+          <property name="binding:*:OnOffType">METER</property>
+        </properties>
+      </channel>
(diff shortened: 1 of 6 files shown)


cdjackson:
Have you added these additions to the database (www.cd-jackson.com)? I updated last night so it should be included if so.

TheNetStriker:
I've just checked the device database and the meter_reset command does not exist in those devices. I just created an Account (Username TheNetStriker) but I don't have the right to change something. Can you give me the right or can you add the changes yourself?

cdjackson:
Thanks - I’ve updated your access.

Cheers
Chris

TheNetStriker:
I've just uploaded the XML's and requested a review, I hope I did everything right. Sadly I only have FGRM222 devices with version 24.24 and 25.25, so I can't upload the file for those devices, but I assume they also support the reset command.

I also just noticed that my 24.24 device doesn't get recognized any more. Do you have any idea what could cause this problem? I've tried to add the device several times in PaperUI. (It worked with the 2.0 release of the binding and I just added the reset command)

cdjackson:
I think this is now all included in the database (thanks), so closing...

PR #409, opened 2017-03-17
Implemented CLOCK/CLOCK_GET

maxberger:
Signed-off-by: Max Berger <max@berger.name>


maxberger:
this is a port of the same commit on the main branch. Note that unlike the main commit this one is only unit tested, not tested with "the real device"

PR #424, opened 2017-03-24
add channel for thermostat_setpoint

berland:
This ZWave unit has a command class that seems to miss a channel - at least I know this hardware can control a thermostat setpoint, but in OpenHAB there is no channel for the thermostat setpoint.

The pull request is an educated guess at what is required - but this is my first OpenHAB contribution - so I do not yet understand if this is everything that is required to get the extra channel properly defined. 

Signed-off-by: Håvard Berland <havardberlandh@gmail.com> (github: berland)


cdjackson:
Please can you add this to the database at www.cd-jackson.com. The database in the binding is automatically generated from this source so it will immediately overwrite any manual changes.

Thanks

PR #542, opened 2017-04-17
Revert to using static logger

cdjackson:


kaikreuzer:
Can you please comment why you do this although it is against the guidelines? I don't think that this is good - if a non-static logger is an issue at some place, it should only be changed there.

cdjackson:
As you saw from the forum there were many problems in many command classes. The quickest way to hopefully get the binding running again was to revert this change. 

As discussed previously this version of the binding has a limited life any way - as soon as the next release is out I will end this branch and use the development version.

cdjackson:
I know what the issue is - for classes that are instantiated through XStream, default constructors and initialisation is not performed - hence if the logger isn't static, it's not instantiated.

I don't plan to fix this for this version but I'll sort this out in the development version - there are already methods in place to handle this - I just need to move the logger instantiation into this method.

kaikreuzer:
Ok, sounds good - thanks @cdjackson!

PR #555, opened 2017-04-23
ZWaveThingHandler: don't assume configuration values are Integers

jvolkman:
I ran into an issue where OH2 was passing the powerlevel_level parameter to the zwave binding as a BigInteger rather than an Integer. The result was a class cast exception. With this change, the values are casted to the more general Number type.


travellingkiwi:
Ping?

cdjackson:
Closing as no longer required following #777

PR #602, opened 2017-06-16
adapted code to refactored interface name "TranslationProvider"

kaikreuzer:
Sorry @cdjackson, I missed to inform you that I just released that change in a new ESH stable.
As ZWave currently breaks the build, we should merge it asap.

Btw, I hope that this code here is really nasty and very much against all good OSGi design principles - you should always avoid putting OSGi services into static fields. I noticed that you did the same in Zigbee as well; it might be worth to consider a refactoring of these parts.

Signed-off-by: Kai Kreuzer <kai@openhab.org>

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/ZWaveBindingConstants.java
+++ b/src/main/java/org/openhab/binding/zwave/ZWaveBindingConstants.java
@@ -11,7 +11,7 @@ package org.openhab.binding.zwave;
 import java.text.MessageFormat;
 import java.util.Set;
 
-import org.eclipse.smarthome.core.i18n.I18nProvider;
+import org.eclipse.smarthome.core.i18n.TranslationProvider;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.openhab.binding.zwave.internal.ZWaveActivator;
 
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks - sorry I missed that the i18 change was merged. Is this change now in OH2 then? I guess it will break people using the development zwave binding if so :(

How would you propose is best to handle this then? I wanted to provide a (static) helper for the i18n services within the binding - how would you suggest is best to refactor (I’m happy to refactor if you have a suggestion :) ).

kaikreuzer:
> Is this change now in OH2 then?

Yes, starting from distro 949, which will be ready any minute.

> How would you propose is best to handle this then?

Well, apart from constants and simple helper methods, nothing should be static, but offered as a service/component. Every service then declares its dependencies and gets their required services injected.
So the best way would be to add a dependency to your handler factories and then pass the translationProvider to the handlers that require it.

cdjackson:
> Is this change now in OH2 then?
> 
> Yes, starting from distro 949, which will be ready any minute.
> 
> 
Any chance we can hold off on this for a few days? I’m not in a position to update things at the moment and this change will break ZWave development binding that a lot of people are using.

PR #655, opened 2017-08-10
Configuration update

easypl:
Added blinds_control, and removed all *1 endpoints.
tested and working now

### DIFF
--- a/ESH-INF/thing/qubino_zmnhca_0_0.xml
+++ b/ESH-INF/thing/qubino_zmnhca_0_0.xml
@@ -11,6 +11,14 @@
 
     <!-- CHANNEL DEFINITIONS -->
     <channels>
+      <channel id="blinds_control" typeId="blinds_control">
+       <label>Blinds Control</label>
+       <properties>
+          <property name="binding:*:PercentType">SWITCH_MULTILEVEL</property>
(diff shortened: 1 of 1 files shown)


cdjackson:
We can't do this as it will now not work for other devices. This should be managed through the database. A separate version of the device should be created so that both the old and new configuration is available. The database can handle multiple versions - otherwise people with the (presumably) older version will not be able to use their device if we change it to the configuration of your device.

easypl:
My module version is:

zwave_class_basic	ROUTING_SLAVE
zwave_class_generic	MULTILEVEL_SWITCH
zwave_frequent	false
zwave_neighbours	17,22,32,35,43,55,58,62,65,67,69,72,80
zwave_version	1.12
zwave_listening	true
zwave_deviceid	2
zwave_nodeid	34
zwave_routing	true
zwave_beaming	true
zwave_class_specific	MOTOR_CONTROL_CLASS_C
zwave_manufacturer	345
zwave_devicetype	3

Please tell me how can I create sub-version?

W dniu 2017-08-10 17:44, Chris Jackson napisał(a):

easypl:
OK, I have created new device, please accept to database

W dniu 2017-08-10 17:44, Chris Jackson napisał(a):

PR #705, opened 2017-10-07
Add missing references for localisation of WallMotes

nullku7:
There are 3 of these 2 button wallmotes
ZW129-A, ZW129-B and ZW129-C
The only difference I can find is the radio frequency.
I have a ZW129-B, and the type:id is 0202:0081
It seems the other 2 are 0002:0081 and 0102:0081

This pull request is simply adding in the 2 additional id's

Having said that, **this is untested**... see this
https://community.openhab.org/t/manufacturer-reference-missing-for-zw129/34874

Andrew Frahn
nullku7@gmail.com

### DIFF
--- a/ESH-INF/thing/aeon_zw129_0_0.xml
+++ b/ESH-INF/thing/aeon_zw129_0_0.xml
@@ -43,7 +43,7 @@
       <property name="vendor">AEON Labs</property>
       <property name="modelId">ZW129</property>
       <property name="manufacturerId">0086</property>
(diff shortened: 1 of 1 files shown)


cdjackson:
This needs to be added to the [database](http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/390) or it will be removed when the database is next updated (ie every few days).

I will update this.

PR #740, opened 2017-11-12
Change actions to use boolean

cdjackson:
Changes actions to use boolean type.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
@@ -252,26 +252,22 @@ public abstract class ZWaveControllerHandler extends BaseBridgeHandler implement
                     continue;
                 }
 
-                if (cfg[1].equals("softreset") && value instanceof BigDecimal
-                        && ((BigDecimal) value).intValue() == ZWaveBindingConstants.ACTION_CHECK_VALUE) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -252,26 +252,22 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
...
                     controller.requestHardReset();
 
-                    value = new BigDecimal(0);
-                } else if (cfg[1].equals("exclude") && value instanceof BigDecimal
-                        && ((BigDecimal) value).intValue() == ZWaveBindingConstants.ACTION_CHECK_VALUE) {
+                    value = false;
+                } else if (cfg[1].equals("exclude") && value instanceof Boolean && ((Boolean) value) == true) {
                     controller.requestRemoveNodesStart();
 
                     value = new BigDecimal(0);
```
openhab-5iver: I would not say I'm familiar with your code, but reviewing your commits helps me to learn it a little. In looking through this commit, this line stood out because of its asymmetry with the other changes. Should this line be replaced with `value = false`?

cdjackson: Good spot - thanks. Missed this one...

PR #748, opened 2017-11-19
Homeseer HSDS100 Updates

dmize:
Fixed channels, default associations, association groups.
I have tested this locally and it is working.

### DIFF
--- a/ESH-INF/thing/homeseer_hsds100_0_0.xml
+++ b/ESH-INF/thing/homeseer_hsds100_0_0.xml
@@ -8,7 +8,7 @@
   <thing-type id="homeseer_hsds100_00_000" listed="false">
     <label>HS-DS100+ Homeseer Z-Wave Door/Window Sensor</label>
     <description><![CDATA[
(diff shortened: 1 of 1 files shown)


cdjackson:
Please make the updates into the database at www.cd-jackson.com and not directly through a PR here.
Thanks.

dmize:
I've updated the database as well.  Although I find it more difficult to update the database and know that it works.  Do you not take PRs anymore?

cdjackson:
The problem with PRs is that they will be overwritten when the database is updated. So I can merge your PR, and it might last a few days or a week or two before being overwritten with the data from the database. For this reason I really prefer to only use exports from the database.

Also, if you do a PR here to the master, then it will not be merged into the development version which uses a slightly different database format - if the data is entered into the database, then we get OH1, and OH2 files all created from the same source.

Why can't you test the database output? I mean, most people don't bother, but if you are testing the XML created by yourself, then it's simple enough to use the database XML - just press the export button.

dmize:
That makes sense.  It took a bit of trial and error to get this device working.  I originally uploaded the xml created when I added it but it didn't work right.  After some research through your interactions with others on the forum I started looking at their database xml and figure out what to do to fix it.  If I had to make each edit on the site and then export it that would have been tedious.  I'll take a look at the export to make sure it aligns with what I have here.  I'm not trying to be a pain, in fact, I'm trying to be helpful.

Thanks for your assistance.

dmize:
Ok, I reviewed the exported XML and it looks good.  Do you want to reject this PR?

cdjackson:
Ok, thanks. I'll be doing a database update tonight so it should go in with that.

Thanks.

Issue #755, opened 2017-11-22
[DEV] Exception during HTTP PUT request for update config


openhab-5iver:
https://community.openhab.org/t/oh2-z-wave-refactoring-and-testing-and-security/21653/1750?u=5iver

This ESH PR will enhance logging to help troubleshoot this issue...
https://github.com/eclipse/smarthome/pull/4403

* OH2 Version: OH 2.2.0 snapshot build 1083
* Binding Version: since 2.2.0.201710012139 

`2017-11-22 07:27:47.869 [ERROR] [marthome.io.rest.core.internal.thing.ThingResource] - Exception during HTTP PUT request for update config at 'things/zwave:device:07cb40a2:node8/config' for thing 'zwave:device:07cb40a2:node8': null`

openhab-5iver:
Could this also be preventing associations being updated, as asked here?

https://community.openhab.org/t/oh2-z-wave-refactoring-and-testing-and-security/21653/2028?u=5iver

openhab-5iver:
Using OH 2.2 snapshot #1140, which includes ESH PR 4403, the exception now looks like this:

```
2017-12-15 22:48:13.434 [ERROR] [marthome.io.rest.core.internal.thing.ThingResource] - Exception during HTTP PUT request for update config at 'things/zwave:device:07cb40a2:node95/config'
java.lang.NullPointerException: null
        at org.openhab.binding.zwave.handler.ZWaveThingHandler.handleConfigurationUpdate(ZWaveThingHandler.java:683) [15:org.openhab.binding.zwave:2.2.0.201712100924]
        at org.eclipse.smarthome.core.thing.internal.ThingRegistryImpl.updateConfiguration(ThingRegistryImpl.java:94) [118:org.eclipse.smarthome.core.thing:0.10.0.b1]
        at org.eclipse.smarthome.io.rest.core.internal.thing.ThingResource.updateConfiguration(ThingResource.java:413) [128:org.eclipse.smarthome.io.rest.core:0.10.0.b1]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
[...]

PR #758, opened 2017-11-29
Avoid NPE during Thing update

evansj:
Compare equality using Objects.equals(Object, Object) which is null-safe.

Possible fix for #755.

Also mentioned here: https://community.openhab.org/t/item-update-results-in-exception-occurred-while-calling-thing-updated-at-thinghandler/34526/2

It’s very hard to add a unit test because the updateNodeProperties() method is private and 144 lines long. Would patches to split this function out be acceptable?

An alternative fix would be to simply do:

```java
update = config.equals(originalConfig);
```

because `org.eclipse.smarthome.config.core.Configuration` implements the `equals` method, but then we wouldn’t be able to have the debug messages showing what changed. Having said that, you’ll only see the first changed key logged anyway because of the `break` statement.

Signed-off-by: Jon Evans <jon.evans@pobox.com> (github: evansj)

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks. Looks good - if you wanted to add test you can use reflection to change visibility in the test.

Yes, if you want to split out the methods, that's also fine. But note that this will be deleted, so it's probably not worth spending too much time on it here - better to spend time on any such refactoring on a PR into the development branch.

Thanks
Chris

evansj:
Thanks Chris,

Please can you explain "this will be deleted"?

I will do some refactoring against the development branch instead, and create some more PRs.

Cheers
Jon

cdjackson:
Sorry - deleted might be the wrong word, but this branch will ‘cease’ and the development branch will become master - it won’t be rebased as they are too different, so changes in master will ultimately be lost (not deleted - wrong word).

Thanks.

evansj:
Makes sense! Thanks.

cdjackson:
Just reopening to kick off Jenkins again to see if this failure is permanent (shouldn't be I think).

openhab-5iver:
Will this be merged into the development branch?

cdjackson:
I looked at this earlier and don’t think it’s an issue in the development branch as there is a null check.

PR #774, opened 2017-12-20
Add "Invert percentage value" option to blinds_lamella & blinds_shutter

mheiss:
*) A new parameter has been added to the blinds_lamella and
blinds_shutter channel definition to invert the percentage value. The
same option is already present on the Rollershutter channel but was
missing in the other channels. Thus the lamella /shutter where moving in
the wrong direction when using this channels.

*) The parameter is evaluated in the proprietary converter used to
control the Fibaro FGR222 shutter.

Signed-off-by: Michael Heiß <michael.heiss@outlook.at>

### DIFF
--- a/ESH-INF/thing/_channels.xml
+++ b/ESH-INF/thing/_channels.xml
@@ -309,6 +309,18 @@
         <label>Blinds Lamella Position</label>
         <description>Sets the blinds lamella position</description>
         <category>Blinds</category>
+        <config-description>
+            <parameter name="config_invert_percent" type="boolean">
+                <label>Invert percentage value</label>
+                <description>Invert the lamella percentage value</description>
+                <default>false</default>
+                <options>
+                    <option value="true">Yes</option>
+                    <option value="false">No</option>
+                </options>
+                <limitToOptions>true</limitToOptions>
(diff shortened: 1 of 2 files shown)


cdjackson:
I guess you have tested this and it's working ok? Otherwise looks fine to me - thanks.
If you confirm it's tested I'll merge and will create a new binding next week when I'm back from holiday.

Cheers
Chris

mheiss:
Can confirm that it's tested. I have multiple FGR222 shutters at home and they are working fine with the new parameter.

Regards,
Michael

kaikreuzer:
> Thus the lamella /shutter where moving in the wrong direction when using this channels.

Just a stupid question: Why do we at all need an option to have it move in the wrong direction? Why not simply fix it in a way that it always moves as expected?

mheiss:
I have multiple shutters at home and it seems that it physically connected some in the wrong way. Thus when i press the UP button it will move down. I would now need to change the physical installation or fix it in the software. As i do not want to again mess around with the installation i was happy to see that i can fix it in the software :D

kaikreuzer:
Ok, that's a good reason then :-D

PR #780, opened 2017-12-29
Database update

cdjackson:


arenk:
#789 The Fibaro Roller Shutter 2 does not work anymore following the update 4 days ago (cannot change the parameters and device is not recognized correctly). The only difference I could see comparing the two versions: Using FGR222 instead of FGRM222

PR #787, opened 2017-12-31
Refactor inclusion code into separate class for readability and testing

### LINKED ISSUE #323: Trying to initialize a new node after cancelling inclusion process
AddNodeMessageClass wrongly initializes a new node when the Inclusion is cancelled via  "doRequestStop" as (at least with razberry controller) the serial message when successfully including a new node and when cancelling is the same -> bytes [0x01, 0x06, 0xNEW_NODE_ID, 0x00].

Haven't tested with RemoveNodeMessageClass but might have the same issue.

cdjackson:
This is mainly a refactoring of the inclusion/exclusion code to remove it from the ```ZWaveController``` class into its own class for testability and readability! It also addresses #323 by not using the node number in the final call - only in the "adding node" response.

A controller priority queue has been added to the transaction manager to ensure that high priority controller messages can be sent immediately. This prevents inclusion commands that might be time critical from getting lost behind other traffic.

Closes #323 


cdjackson:
Closes #788

cdjackson:
@openhab-5iver let me know if you don't think this is good for prime time, otherwise I'll merge tonight. It has a few small changes from the last version you ran, but nothing too major and I think it's worth getting this into the system ;)

PR #792, opened 2018-01-05
Workaround for buggy IRIS CT101 thermostat

marcusb:
The IRIS version of the Radio Thermostat CT101 has a buggy Zwave
implementation.

Original patch by Jonathan Schmidt.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
@@ -1283,12 +1283,15 @@ public class ZWaveNode {
             }
 
             // Check that the length is long enough for the encapsulated command to be included
-            if (payload.getCommandClassCommand() == 6 && payload.getPayloadLength() > 4) {
+            if (payload.getCommandClassCommand() == 6 && payload.getPayloadLength() > 4
+                    && !hasMultiInstanceAsMultiChannelQuirk()) {
                 // MULTI_INSTANCE_ENCAP
                 endpointNumber = payload.getPayloadByte(2);
 
                 payload = new ZWaveCommandClassPayload(payload, 3);
-            } else if (payload.getCommandClassCommand() == 13 && payload.getPayloadLength() > 5) {
+            } else if ((payload.getCommandClassCommand() == 13
+                    || payload.getCommandClassCommand() == 6 && hasMultiInstanceAsMultiChannelQuirk())
+                    && payload.getPayloadLength() > 5) {
                 // MULTI_CHANNEL_ENCAP
                 endpointNumber = multichannelCommandClass.getSourceEndpoint(payload);
                 payload = new ZWaveCommandClassPayload(payload, 4);
@@ -1528,5 +1531,10 @@ public class ZWaveNode {
             timerTask.cancel();
         }
         timerTask = null;
+    }
+
+    // The IRIS Radio Thermostat CT-101 has a bug where it sends an incorrect command class code.
+    private boolean hasMultiInstanceAsMultiChannelQuirk() {
+        return manufacturer == 0x98 && deviceType == 0x6501 && deviceId == 0xc;
     }
 }


marcusb:
Although this may be a hacky workaround, it makes the thermostat work so I'm making a PR for consideration, and so users can find the patch more easily.

Credits to [jschmidt](https://community.openhab.org/t/ct101-thermostat-not-supported-in-openhab2-z-wave-binding/14384/33?u=marcusb).

cdjackson:
I'm probably happy to merge this since the master version will be end of life-ed very soon so a hacky solution is probably ok.

Note though that this [needs to be signed](https://docs.openhab.org/developers/contributing/contributing#sign-your-work) - especially if it's not your work.

marcusb:
I'll have to defer the signing to the original author.

jeschmidt:
I am happy to sign it, but not sure how to add this to the commit message.  I have reviewed the Developer Certificate of Origin and can attest to the terms.  If it helps, you may add this line:

Signed-off-by: Jonathan Schmidt <jon-github@jschmidt.ca> (github: jeschmidt)

marcusb:
Sign-offs added to commit message.

jeschmidt:
Good to merge now?  Let me know if there's anything else I can do to help.

jersoe:
This seems to address the issue I'm having. Do I understand it correctly that this has not been merged, and therefor is not in the stable or nightly releases. I tried downloading this branch and building with mvn, but it fails two tests. Do I just ignore those? Do I install the resulting jar in the addons directory?

jeschmidt:
I don't think it has been merged yet.  @cdjackson can you confirm?

I'm not sure about the two failed tests.  I'm still using the zwave binding built from this branch so I can use my CT101's.  Very stable, no issues with months of runtime, but of course I don't have the benefit of the more recent changes in the latest build either.  I'd love to see them merged together :)

jersoe:
I've forked the master and reapplied the patch, since the master has 50+ commits that weren't in this one. I then repackaged using Maven and it compiled without failing tests. Fork at https://github.com/jersoe/org.openhab.binding.zwave and for anyone having trouble compiling, I've temporary put the jar at https://www.jersoe.com/org.openhab.binding.zwave-2.4.0-SNAPSHOT.jar. I'll try to figure out today how to install this in OpenHAB.

cdjackson:
In the coming week or two this branch will become obsolete.

jersoe:
What will it be replaced with, and will the new version address the Iris CS-101 issue?

cdjackson:
The development version will take over, and no, it won’t address this as there’s currently no nice solution available. The current hack is really not so nice (as discussed above and elsewhere).

volfan6415:
So then will people using CT101's be out of luck?

cdjackson:
For now, the CT101s with the major flaw will not work with the dev binding - correct. Those that don't have the bug should work fine. 

Personally, I would have suggested to send it back to the manufacturer for replacement as it's a very major bug and very non-compliant to the ZWave standard (and I suspect will not work with many controllers). If we can find a nice way to work around this then we can look at adding it, but this bug is really in the low level of the protocol layers and it's a horrible kludge.

jersoe:
Reapplying the patch to the 2.4.0 snapshot did the trick and I got it working. I see your point, but on the other hand I think it would be a shame to have OpenHAB not support these amazing cheap thermostats. Home Automation for example correctly recognized it out of the box as Iris CT-101 and it just works. How many users will find this, and start downloading branches and recompiling and all? I'd like to find a permanent solution. If I find some time this week I'll try to read through the code to see if I can come up with something...

cdjackson:
> Reapplying the patch to the 2.4.0 snapshot did the trick and I got it working.

I'd be VERY surprised if this was the latest version since it has undergone heavy refactoring?

> I think it would be a shame to have OpenHAB not support these amazing cheap thermostat

I guess they are "amazingly cheap" because they don't work on most systems? 

While I agree that it would be nice to support them, I also don't want to compromise the binding to support devices that have such a major flaw, so we need to try and find a nice way to add this. Maybe we'll have to go with something hacky, but I really don't like this.

jeschmidt:
While I don't know for sure, I do suspect the low price directly relates to the communication bug.  I think the mfr. made a *lot* of these as I keep seeing them posted on different retail websites for excellent prices.  I suspect they cannot be fixed because the same brand sells a different version of the thermostat without the bug for 4x or higher price.  Supporting them in OH is a benefit to the OH community as it can save us a lot of money (until supply runs out!).  That was my original rationale for spending the time to hack the binding to make them work.

Chris, way back then you mentioned that there was a more acceptable mechanism for implementing this.  I had said I would take a look but my infant daughter inverted some priorities...  Anyway, I recall it being something about the split between the database and the source code, and that the device-specific workarounds are kept in the database somewhere.  If this could be re-written to look more like the other workarounds, maybe it would be more acceptable to merge to mainline.  Is that someone can look into?  (@jersoe ?)

Philosophically I would hate to see OH take the stance that we only support 100% compliant zwave hardware.  I don't think that's your point and certainly there are other device workarounds in the binding/DB already.  I'm not sure I understand your point perfectly but I think this is more about implementation ugliness than about supporting a non-compliant device.

One last thought on the hack itself.  It's embedded in the error-handling path of the protocol parsing.  You don't get there unless there's a violation already.  I think of it as something like this:
- Hey, that packet's not properly formed, we should flag an error!
- Oh wait, is it from this device which knowingly makes this mistake?
- Right, it's a CT101.  I know what it meant instead.  Fix it, and continue parsing...
So the chance for this hack to affect something it wasn't targeting is pretty low.  The other device would need to make the same protocol mistake that CT101 does, and then have the same vendor/device IDs.

jwbusch:
Any word on when this might make it in? And/or could someone share a private .jar to test? (I tried building it, but I'm not familiar with Java and hit some build errors). I also bought some CT101s before doing more research (Assumed Z-Wave would be compatible), so I also would love to get these working.

jwbusch:
Managed to get it to build and can confirm it seems to work perfectly for my CT101s. 

Figure I'll share instructions for building on Windows for people who need a one-off .jar to test on their system. I saw some other people having trouble building it and can't really think of a better place to put this 🙂. 

I use https://scoop.sh on Windows as a package manager, so I'm assuming that you'll also have it installed (It's basically apt-get/homebrew for Windows -> makes installing dev tools/utilities simple). 

- Get Maven and Oracle JDK for building the jar
```
scoop add bucket java
scoop install oraclejdk8
scoop install maven 
```

- Clone and build w/ Maven 
```
git clone https://github.com/marcusb/org.openhab.binding.zwave.git
cd org.openhab.binding.zwave 
git checkout mbetter/ct101
mvn package
```
After a few minutes (mostly downloading dependencies) there you should see `BUILD SUCCESS` and you can look inside the `org.openhab.binding.zwave/target` dir for `org.openhab.binding.zwave-2.4.0-SNAPSHOT.jar`. Remove your original Z-Wave binding (via PaperUI or similar), drop the jar in your `%OPENHAB%/addons` folder, and restart openHAB. It took my CT101s a few minutes to start reporting any data.

jeschmidt:
I just checked the patch from @marcusb again.  He already pulled the workaround out into a flag function (hasMultiInstanceAsMultiChannelQuirk), which is an improvement from the earlier (ugly) implementation I made.  Can we add this quirk flag to the device database (enabled only for CT101 initially), and that function can look it up from the DB instead of comparing against the manufacturer/device ID in the binding (Java) code itself?  @cdjackson would that make the workaround more acceptable?

marcusb:
@cdjackson How does this look?

marcusb:
For those who want a patch conveniently applied to 2.5.x, there is a branch [here](https://github.com/marcusb/org.openhab.binding.zwave/tree/mbetter/ct101-2.5)

cdjackson:
Replaced by #1344

PR #838, opened 2018-02-21
Use correct thermostat setpoint type for Eurotronics Spirit (Fixes #837)

### LINKED ISSUE #837: Eurotronic Spirit thermostat: Incorrect thermostat setpoint type
* OH2 Version: 2.2.0
* Binding Version: 2.3.0.201802081624

If the problem is associated with a device, please provide the following -:

* Device name: Eurotronic Spirit
* Link to [database](http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list): http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/710

The "energy heat" thermostat setpoint is linked to the wrong type ("FURNACE" instead of "HEATING_ICON"). This patch adjusts this and also renames the setpoint to match the Spirit "energy heat" terminology. I am not sure if the latter is desired or correct, please let me know.

[Others](https://community.openhab.org/t/problems-with-eutronic-spirit-z-wave-plus-thermostatic-valve/39397) have ran into this problem as well and I can confirm that this patch allows me to set the "energy heat" setpoint correctly.

RoyJacobs:
This is a fix for #837 

Signed-off-by: Roy Jacobs roy.jacobs@gmail.com (github: RoyJacobs)

### DIFF
--- a/ESH-INF/thing/eurotronic_spirit_0_0.xml
+++ b/ESH-INF/thing/eurotronic_spirit_0_0.xml
@@ -45,10 +45,10 @@ Thermostatic Valve<br /><h1>Overview</h1><h1>Intelligenter Energiesparregler fü
           <property name="binding:*:DecimalType">THERMOSTAT_SETPOINT;type=HEATING</property>
         </properties>
       </channel>
-      <channel id="thermostat_setpoint_furnace" typeId="thermostat_setpoint">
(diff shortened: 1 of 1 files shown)


cdjackson:
Please update the database at www.cd-jackson.com. Any PR accepted here will simply get overwritten when the database is updated.
Also, you can't simply change the channel name if the channel is not supported.

PR #850, opened 2018-03-10
[dev] Database update - RESYNC

cdjackson:


mhilbush:
I was a couple minutes too late with my change to FGD212...

cdjackson:
Yeah - I just saw them. I’ll get them in here before I merge this, but the merge into master will have to wait till the next round…

And thanks for doing this :)

mhilbush:
I guess I need to start waking up earlier.  LOL

cdjackson:
Speak to @openhab-5iver - he seems to have perfected being always awake ;)

cdjackson:
Updated with the FGD212 - is that all your changes done for now @mhilbush ?

mhilbush:
Yes.

cdjackson:
@mhilbush @openhab-5iver what do you think? I've restructured the database into separate folders per manufacturer as we're heading towards 800 entries now and it's just too unwieldy (IMHO) in one folder.

I also removed the manufacturer off the filename itself.

Happy to hear any thoughts before I merge ;)

openhab-5iver:
I can't think of any reason not to do this. Do you have concerns about it?

cdjackson:
No - no concerns. I initially thought that everything needed to be in one folder for ESH to find them, but I checked this morning and it does recurse the folders.  I did a quick restart and it seems to load the files fine.

openhab-5iver:
If you'd like, send me the jar and I can smoke test before you post it.

cdjackson:
Thanks. I’ve put it in my usual download folder - filename is org.openhab.binding.zwave_2.3.0.201803111303.jar.

mhilbush:
I'm a little late joining the conversation, but I agree that it makes sense to organize into individual folders per manufacturer. I didn't realize it was up to 800 entries. Seems like new devices are coming out almost daily...

On a related topic, I also agree with @openhab-5iver that the ability to see the history of database updates would be great. I currently do this by sorting through the commits, which works, but is less than ideal.

cdjackson:
The easiest way to see the history is to go to the XML file you’re interested in, and click on history, or click on blame. History will give you the list of changes, and blame will show each line that changed with a date (but it won’t show removals). 

I agree that something a bit nicer would be, well, nicer, but it’s also quite a bit more work. I could feed the whole thing in to a git repo and record changes every time someone changes anything, but that might slow things down even more. I’ll have a think some more...

Currently there are 713 devices exported, but yes, it’s going up quite fast at the moment!

openhab-5iver:
Restarted OH with the 2.3.0.201803111303 jar, deleted 119 Things, and there's nothing out of the ordinary in the log. Everything seems to function as it did, so looks good to me.

PR #852, opened 2018-03-10
Fix NPE in association update

maxberger:
Signed-off-by: Max Berger <max@berger.name>

---

For controllers which do not have any associations yet the "currentAssociation" can be null. This leads to an NPE such as the one below. This  happend on an un-associated WALL-C switch, with this fix I can add an association.

``22:39:23.770 [ERROR] [est.core.internal.thing.ThingResource] - Exception during HTTP PUT request for update config at 'things/zwave:device:f851f020:node30/config'
java.lang.NullPointerException: null
        at org.openhab.binding.zwave.handler.ZWaveThingHandler.handleConfigurationUpdate(ZWaveThingHandler.java:623) [252:org.openhab.binding.zwave:2.3.0.201803101457]
        at org.eclipse.smarthome.core.thing.internal.ThingRegistryImpl.updateConfiguration(ThingRegistryImpl.java:94) [117:org.eclipse.smarthome.core.thing:0.10.0.b1]
        at org.eclipse.smarthome.io.rest.core.internal.thing.ThingResource.updateConfiguration(ThingResource.java:413) [127:org.eclipse.smarthome.io.rest.core:0.10.0.b1]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:497) ~[?:?]
``

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -618,6 +618,7 @@ public class ZWaveThingHandler extends ConfigStatusThingHandler implements ZWave
                 }
 
                 ZWaveAssociationGroup currentMembers = node.getAssociationGroup(groupIndex);
+                if (currentMembers == null) currentMembers = new ZWaveAssociationGroup(groupIndex);
                 ZWaveAssociationGroup newMembers = new ZWaveAssociationGroup(groupIndex);
 
                 int totalMembers = currentMembers.getAssociationCnt();


cdjackson:
Can you advise what the situation is that means the association group is not found in the first place? Is there an error in the database or is the device not initialised, or something else?

The implementation in the dev binding is different - it would be good to understand what the root cause of this is rather than just ignoring it and it is probably good to synchronise the two implementations.

Also, if we merge this, it will need to be formatted correctly as per Eclipse.

baumgm:
I did a complete reset of my Fibaro Motion Sensor and Door Sensor. After that i tried to include them again in the network. But i can not set any association group because of the same error above. The devices worked correctly before the reset.

Right now the devices are useless.

maxberger:
@baumg Any chance you could test my patch?

I had similar problems, and read chris's response, but haven't been able to
follow up with it yet.

In my case it was the wall-c switch, also newly added, so no associations
where set.

Max

baumgm <notifications@github.com> schrieb am Di., 13. März 2018, 21:57:

cdjackson:
The other option is to use the development version which has this issue fixed already. When I get some time this will be merged into master.

cdjackson:
As development version is now merged, this is not applicable.

PR #864, opened 2018-04-01
Update qubino_zmnhtd_0_0.xml

HerzScheisse:
Typo...

### DIFF
--- a/ESH-INF/thing/qubino_zmnhtd_0_0.xml
+++ b/ESH-INF/thing/qubino_zmnhtd_0_0.xml
@@ -26,7 +26,7 @@ Smart Meter<br /><h1>Overview</h1><p>This Z-Wave module is used for energy measu
         </properties>
       </channel>
       <channel id="meter_voltage" typeId="meter_voltage">
-        <label>Votage</label>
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks, but please update the database directly as this will be overwritten otherwise.

http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-database-guide

HerzScheisse:
thanks Chris, yeah i tried that  and also read the guide on your page... but "To upload an XML, select the Update Device option from the tools menu in the top right of the device summary page." i dont have that "tool" menu...

cdjackson:
I've updated your access, so you should see it now.
Note that uploading the XML won't fix the typo - the XML that gets uploaded here is not the one that is in the database - it's the one that OH saves once it has read all the data from the device so it doesn't contain any descriptions etc (basically, it populates all the endpoint related stuff at the bottom, and device level stuff at the top).

Just edit the database directly - you should now see a lot of edit buttons once logged in...

Thanks.

HerzScheisse:
Thanks... Will have a look :)

PR #902, opened 2018-05-11
Support discovery of USB dongles (closes #900)

### LINKED ISSUE #900: Support discovery of USB dongles
Hi @cdjackson!

I know that you are busy, but it would be awesome if USB dongle discovery could be added to the ZWave binding, just like it is now in ZigBee binding. Perhaps @hsudbrock could give a helping hand here, as he has develop the USB discovery API :wink: I can try to implement it myself, but I'm not sure if I should push it to master or development branch. How far are we from merging both branches? I suppose that it's imposible to get it for openHAB 2.3 release.

Anyway, here you can find the properties of the two Aeon ZWave sticks:

- ZStick Gen5 -> Vendor_id: 0x0658, Model_id/Product_id: 0x0200
- ZStick Gen2 -> Vendor_id: 0x10c4, Model_id/Product_id: 0xea60

Thank you for the amazing work and best regards,

Aitor

bodiroga:
The PR adds support for discovering two Z-Wave USB dongles:

- Aeotec ZStick Gen2 (idVendor: 0x10c4, idProduct: 0xea60)
- Aeotec ZStick Gen5 (idVendor: 0x0658, idProduct: 0x0200)

Problems found:

- The ZStick Gen5 doesn't provide any serialNumber when it is discovered, so the idProduct number is used as the ThingID.
- Should I create a new "src/test/java/org.openhab.binding.zwave.discovery" for the tests?

I have test it with both dongles and it works fine in my development machine, so many thanks to @hsudbrock for his work!

Best regards,

Aitor


cdjackson:
Thanks @bodiroga - can you please sign the PR and commit please?  Other than that I don't have any comments - looks fine.

@hsudbrock - any comments?

bodiroga:
All commits squashed into a single one and the PR is now signed off. 

Thank you for all your work @cdjackson! :+1:

hsudbrock:
Looks good to me as well, thanks @bodiroga.

Just as thought, if you like to add it now or another day: One thing that could be improved (and which I plan to improve for the ZigBee binding as well) is to not only populate those properties in the discovery result that are sufficient to configure the bridge, but to also populate some of the ESH standard properties that can be added optionally as metadata to things. With  these I mean the properties whose names are defined as constants in the `Thing` interface, e.g., `serialNumber`, `modelId`, or `vendor`.

Another thought: By using the product ID as thing UID, you of course prevent having two Z-Wave USB dongles (of the same type) being used side by side. To me this sounds like an edge case; if you want to consider this anyway, you could think about using, e.g., the current timestamp in millis as thing UID (or something sufficiently random), as it is just used UID.

I see both points above as optional improvements, which don't need to go into the discovery, or could be added later.

bodiroga:
Here I am again guys!

Another possibility: what do you think @cdjackson if I establish a serial connection with the controller when the usb device is discovered to get the ZWave homeId and use it as the thingId? The homeId should be unique per controller, so multiple controllers could be attached to the computer at the same time. I could also read the manufacturerId and the deviceId to set the correct label for each thing and don't rely on the serial idVendor and idProduct.

The discovery process would be slower (and I'm not sure if ESH discourages to establish serial connections when implementing the UsbSerialDiscoveryParticipant), but IMO there is a clear benefit in doing so.

Best regards,

Aitor

hsudbrock:
> wouldn't that behave incorrectly if the user disconnects and connects the usb device? The UsbSerialDiscoveryParticipant would create a new thing with the new random thing ID and it would be added to Inbox, confusing the final user.

The `org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcesor` should prevent this, as the `representationProperty` is set to the serial port. (Unless the `AutomaticInboxProcessor` is set to not auto-ignore inbox entries, but  that is not the default.)

As I understand it, the `AutomaticInboxProcessor` will ignore an inbox entry, if there is already a thing of the same thing type that has the same value of the representation property, i.e., in this case, a Z-Wave dongle that is configured for the same serial port. So only if, after unplugging and plugging the USB dongle, it gets another serial port (from udev or mdev or whoever is responsible for creating the device nodes on the Linux system), then a new discovery result would be added. But in this case, the user would have a problem with the existing thing anyway, as it would be configured for the wrong serial port then.

bodiroga:
Many thanks for the explanation @hsudbrock, I really appreciate it :+1:  Just one question: how do you know that the serialPort is defined as the representationProperty of the Thing? The reason why I ask it is because you are right, if you disconnect the dongle from the usb port and connect it again, usually Linux assigns a different port to the device, so changing the representationProperty will help here. But anyway, the binding is not able to reconnect to the controller after unplugging it from the usb port (even if the port didn't change), so we are getting very off-topic here, sorry.

@cdjackson, is there anything that you would like me to change in this PR? :wink:

cdjackson:
Sorry - I was away over the weekend so just catching up...

> Another possibility: what do you think @cdjackson if I establish a serial connection with the controller when the usb device is discovered to get the ZWave homeId and use it as the thingId? 

I don't know that I like this - especially with the discussion about serial ports changing... This adds a lot of complexity to the discovery - you would need to duplicate some of the binding to read out the home ID etc. This isn't impossible of course, but I would fear that opening the serial port and trying to read data from the device would conflict with the binding if something went wrong with this detection, and that would block the actual binding working.

Maybe I'm being too pessimistic, and  I'm not sure how likely this is, but to me I'm not sure if the complexity is necessary and if it might cause more problems than it really solves? What do you think? I'm just worried about having to answer lots of "the binding won't start because the serial port is blocked" sort of question. ;)

> The homeId should be unique per controller

The homeID is not 100% fixed for a dongle, although it's pretty unlikely to change without a reset (but it is possible), We should just be careful with this assumption if we're really relying on it.

> @cdjackson, is there anything that you would like me to change in this PR? wink

I'll take another look over this later, but I did quickly review it on Friday and I didn't really spot any issues.

bodiroga:
Sorry for being so annoying, but I think that I'm going to make some changes in the PR. I have tried with different Z-Wave Plus dongles and all of them are being identified with the same vendor and product id as the Aeotec ZStick Gen5. IMO it will be better if I just implement:

Z-Wave USB dongle (idVendor: 0x10c4, idProduct: 0xea60)
Z-Wave Plus USB dongle (idVendor: 0x0658, idProduct: 0x0200)

With the information provided by the USB discovery service we are not able to identify each particular controller and, as establishing a serial connection with the device is not a great idea, our best option is to name the Thing with a generic name.

I will change the code accordingly, thanks for your patience!

bodiroga:
PR updated! :+1:

cdjackson:
Thanks @bodiroga . This looks fine to me - I will merge this in a couple of days. The only thing stopping me now is waiting for 2.3 to be released. I know that this goes into the dev branch, but it will require people to have 2.3 snapshot or final due to the USB features, and when we added this on ZigBee I got a load of questions from people who had older runtimes ;)

bodiroga:
Hi @cdjackson!

Doesn't the UoM support that you introduced force the user to use the 2.3 version of openHAB? I think so. Anyway, openHAB 2.3 is officially released now: https://www.openhab.org/blog/2018-05-27-openhab-2-3-release.html :wink:

cdjackson:
Probably, but I forget when the different things were merged into 2.2. I think UoM was merged a long time ago where the USB mods were more recent. I just want to reduce the number of complaints.

Yes, I know that 2.3 is available, but again I just want to give people a few days to sort things out - just to reduce the number of complaints I get...

bodiroga:
Then we can wait a couple of weeks until people jump to the new release, don't worry!

PR #912, opened 2018-05-22
Database update + thing documentation

cdjackson:


cdjackson:
@BClark09 do you mind taking a quick look at this please.  The change to look at is that I've refactored the thing documents slightly - as with the database files themselves, the md files are now in separate folders per manufacturer to avoid having 750 files in a single folder!  I'm assuming that the documentation scripts will handle this (ie it will handle sub-folders of the /doc folder, and all files don't need to be in the /doc folder?).

I'd like to get this merged for 2.3 if possible as it should align the database with the documentation (always good I think ;) ).

BClark09:
Hi @cdjackson,

It looks good to me, but I'm not sure if there's any thing that needs to be done for the new website. Perhaps @ghys might be better to answer?

ghys:
Ah yes - I tried including those pages on the new website, however, while it works, I left them excluded in the end because they triple the build time as well as the size of the client-side router's array, search index etc. to IMHO unacceptable levels.

I don't think it's the best forward to generate them all as static pages like this considering the way VuePress (the new site's framework) works. However, it would be feasible to build a new layout or component constructing the page dynamically from the thing-type xml, which would be fetched on request from the id in the query string - like `thing.html?id=act_45602_00_000`.

In the meantime, `things.md` can still be included, but I would appreciate it if you could replace the hyperlinks to the generated Markdown files with direct links to http://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/{id}.

Thanks!

cdjackson:
That’s a shame. I didn’t want to just link to the database as it requires more knowledge to interpret. The idea with the documentation is it is more user oriented and it would really be a shame if we can’t support the Zwave thing documentation :(

If it’s really a problem though I will remove it and we just have to let people work it out themselves.

ghys:
No worries, I recognize the value of these documents, and they will be back, probably with a dynamic page like I suggested (since it's all generated info) but since the website rolls out on Friday, it won't be possible to have it ready by then.

ghys:
Just to be clear, docs.openhab.org is not going away on Friday so they would still be there too.

cdjackson:
Ok, so I’m a little unclear what to do with this PR. Currently, as we agreed in January, the files are already in the repo, so all this PR does is to update them and move them into subfolders. Is it ok to merge this? If not I need to separate this out as at least part of this needs to get merged for the 2.3 release.

kaikreuzer:
> (Since the line I added does not work on Mac, I need to revise it anyhow. )

 FTR: [This is](https://github.com/openhab/openhab-docs/pull/678#issuecomment-391006571) what @BClark09 refers to.

>  it would be feasible to build a new layout or component constructing the page dynamically from the thing-type xml

This is very much related to https://github.com/eclipse/smarthome/pull/1853 and https://github.com/eclipse/smarthome/issues/5571.

As @cdjackson says, so far our approach was to generated md from the XMLs and have the result included in the website. The good thing about that approach would be that we do not depend on the XMLs, but that we could create the mds also by other means (as for Z-Wave).

If the content is simply too much, would it be possible to exclude them from the search index and just have them "sit there" and be available through a link from the main readme page?

doc/act/zdw120_0_0.md
```diff
@@ -0,0 +1,270 @@
+---
+layout: documentation
+title: ZDW120 - ZWave
+---
+
+{% include base.html %}
+
+# ZDW120 Two Wire Wall Dimer
```
kaikreuzer: Could you adapt your generation so that it makes sure to always have an empty line after every header? The current files will break the Jekyll renderer...

cdjackson:
Will do - thanks.

kaikreuzer:
@ghys As all links work as expected on this preview website, I assume that https://github.com/openhab/openhab-docs/pull/678 is actually not required at all?

ghys:
> @ghys As all links work as expected on this preview website, I assume that openhab/openhab-docs#678 is actually not required at all?

No.

ghys:
> Does build time matter?

I just read about deploy contexts here https://www.netlify.com/docs/continuous-deployment/#deploy-contexts

If we end up compiling the ZWave things documents, I think it would be feasible to exclude them from preview deploys (when reviewing a PR) via an environment variable or a commandline switch to speed things up, and only enable them in production.
This doesn't help the pages index size but that's still something to consider.

ghys:
@cdjackson I just thought about another solution - simply render the markdown client-side on a special page, and include links in the search results for ZWave things.
I created https://github.com/openhab/website/pull/24 for this.

Preview: https://deploy-preview-24--openhab-website.netlify.com/addons/

The idea is, you search for a thing (try "smoke" or "door") and if it's a ZWave thing, clicking on the name will lead you to a page able to render the doc dynamically from the markdown source.
It seems to work, and addresses my concerns - and no need for the big table in things.md anymore IMHO.

ghys:
Btw @cdjackson - the markdown file names were wrong, you stripped the leading zeros in the things UIDs - i.e. 45602_0_0.md instead of act_45602_00_000.

It's fine, I handled them but if you intent to fix them, please let me know so I can adapt my code. Thanks :)

cdjackson:
I definitely like the search - it’s really good. However, I do think that we should have the static links as well - or alternatively, we should remove the links so at least they aren’t broken (ie https://deploy-preview-24--openhab-website.netlify.com/addons/bindings/zwave/doc/things.html <https://deploy-preview-24--openhab-website.netlify.com/addons/bindings/zwave/doc/things.html> doesn’t work).  Personally, I would prefer to keep the static links so that people looking for a list of supported zwave devices have somewhere to look - this is a reasonably common question and it would be good if the user had such a list.

I will fix the links - sorry for this. This came about when I moved the files into the separate folders…  I plan to do an update of this tonight - I’ll ping you when I do.

Thanks for the work - it does look good.

ghys:
Keeping things.md is fine :)
If you want to adapt the links in the table, I think (hope) this will work:

`[./thing.html?thingTypeUID=<uid>](<model>)` (replace the placeholders)

At some point I'd like to have a full-text search for the search box in the top bar, so it would likely find the table when typing a model of manufacturer name.

cdjackson:
Hmmm - I updated the links here, but didn’t account for your new format. @ysc - I wonder if it’s not better to make this change in the doc script? If I change this in the source, then the links won’t work in GitHub - this was something we had tried to maintain in the previous discussions. I think if I make this change to the doc source, then it makes an assumption about the final implementation that may also not be correct for all usages (ie if someone uses this for other purposes other than directly in OH) so I think it should be the openHAB doc script that makes this change?

What do you think?

ghys:
You're absolutely right - keep the links so they match the folder hierarchy and I'll deal with the website-friendly links :)

cdjackson:
Ok, sounds good - thanks.  I’ll merge this then - if you get a chance to create another preview I’ll take a look to make sure things look ok. I might make a few small changes, but I don’t want to spend too much time on it right now and I already think it’s a huge step from previous versions.

ghys:
In particular I saw pictures of devices in the database, this could be valuable in the docs if we find a way to include them somehow...

@kaikreuzer could you update the external resources in openhab-docs please?

cdjackson:
Yes - the images were something I was planning to include, but it needs a little work on my side first. I want to process them so they are all the same size, and in cases where there are multiple images in the database, I need a way to specify which one is added. I probably won’t include images in the repo, but will link out direct to the database - just to keep sizes etc down.

kaikreuzer:
> @kaikreuzer could you update the external resources in openhab-docs please?

Done!

ghys:
Thanks!

@cdjackson could you check again https://deploy-preview-24--openhab-website.netlify.com/addons/bindings/zwave/doc/things.html
and approve https://github.com/openhab/website/pull/24 so I can merge?

PR #915, opened 2018-05-26
Update documentation to add images and documentation links, and battery channel

cdjackson:
@ghys this mostly fixes a little formatting, and adds some missing channels that don't come through user entry in the database.

There is one significant change though - I've added images. These are just links to the database, so there are no images in the repo. The database resizes the image to 300x300 so everything should be uniform.

The image links are provided at html rather than md links since md only seems to allow relative links.

Two questions - does the cross site image link cause any problems, and is the html link here ok? If not, I can easily remove these links but I think it's a nice addition if possible (and I think you also wanted to see these added based on your comment yesterday?).


ghys:
Hi Chris,

> does the cross site image link cause any problems, and is the html link here ok?

Unfortunately, as much as I would like those images on the pages, it's a bit problematic. When you link non-HTTPS images from an HTTPS site, the browser "taints" the page as "not fully secure"; this looks bad, and since it's a single-page application it remains even when navigating away. There are other places (example: https://www.openhab.org/addons/bindings/homematic1/) where this happens, and will hopefully be fixed.

Is there any chance you could enable HTTPS on your site? Then it would be good for me. Another solution could be to move them to GitHub since we can always link to https://raw.githubusercontent.com/openhab/...

cdjackson:
Thanks Yannick,
I prefer not to pur all the images on GH for a couple of reasons. I’ve enabled HTTPS on my system and updated the certificate so hopefully this works. I’ll update the PR to change the links to https:// and you can give it a try…

I must say that the site looks great, and I really like the dynamic thing search - it makes the system feel a lot more professional :)

Cheers
Chris

cdjackson:
I've updated the image and manual links to use https - let me know if that's ok now...

Thanks
Chris

ghys:
Excellent!
I've temporarily changed the component to fetch from this PR's branch to test, and there's one problem left:
the client-side renderer doesn't understand raw HTML in the markdown.

image

There's probably an option somewhere, but could you change the images to Markdown syntax instead, like this:
```
![DSB45 product image](https://www.cd-jackson.com/zwave_device_uploads/92/92_default.jpg)
```
Then it will look good:

image

Thanks!

cdjackson:
Thanks - I’ll change to using the MD links. The reason I didn’t use them before is that some renderers I tried didn’t like absolute links...

cdjackson:
Ok, now updated with MD style image links - if you're happy with this, I'll merge.

Thanks.

ghys:
LGTM now :)
Note that it uses the files in openhab-docs, not in this repo (could be changed if necessary) so it will only appear on the website when those are updated.

kaikreuzer:
> you mean that there’s a script that pulls the docs from here across to openhab-docs

This script is me for the moment, doing that manually. Definitely time to automate it some day...
Just pushed an update to openhab-docs, so all should be up to date now.

cdjackson:
Ok, understood. Thanks both.

PR #940, opened 2018-06-30
Fix for issue with EzMultiPli color output (#939)

hexamer:
Fixed as suggested by @cdjackson: refactor color commands to sepearate sending of R,G,B and CW,WW

Signed-off-by: Aaron Hexamer <hexamer@comcast.net>

PR #942, opened 2018-07-01
Refactor color converter

### LINKED ISSUE #939: Color command processing incompatible with EzMultiPli
This device only has discrete red, green, and blue LEDs (no white LEDs), but also uses the CW color state to turn on or off all LEDs to simulate white.  So if a color switch command is sent with a CW component (after the R/G/B components), as is the case with the current binding, there is no way to get any state other than off or white.

Per discussion in this thread, the proper fix is to refactor the color commands such that RGB and CW/WW are sent separately: https://community.openhab.org/t/problems-commanding-color-of-ezmultipli/46683

* OH2 Version: 2.3.0
* Binding Version: 2.3.0

* Device name: EzMultiPli 2.0
* Device version: 2.2
* Link to [database]:https://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/268

cdjackson:
closes #939 
replaces #940 

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveColorConverter.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveColorConverter.java
@@ -129,8 +129,8 @@ public class ZWaveColorConverter extends ZWaveCommandClassConverter {
                 colors.put(ZWaveColorType.BLUE, 0);
             } else {
                 colors.put(ZWaveColorType.RED, scaleColor(color.getRed()));
-                colors.put(ZWaveColorType.GREEN, scaleColor(color.getRed()));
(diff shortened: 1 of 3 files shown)


cdjackson:
@hexamer please retest this - I had to make some changes as at least the Aeon bulb needs to set the whites to 0. I've done this by internally checking the supported colors so I think it should still work fine for your device.

hexamer:
Retested and confirmed it works with EzMultiPli
This is thankfully due to the fact that EzMultiPli is only reporting R,G,B in SWITCH_COLOR_SUPPORTED_REPORT, which I confirmed in the log, even though it does (incorrectly) support CW commands too.

cdjackson:
Thanks.

PR #950, opened 2018-07-03
Delete the XML file when node is excluded or thing is removed

### LINKED ISSUE #948: [DEV] Deleting node Thing also should delete node.xml file
When a zwave node Thing is deleted, the binding also should delete the associated node.xml file.

cdjackson:
Closes #948 

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
@@ -778,4 +778,13 @@ public abstract class ZWaveControllerHandler extends BaseBridgeHandler implement
         getThing().setProperty(ZWaveBindingConstants.PROPERTY_NEIGHBOURS, neighbours);
         getThing().setProperty(ZWaveBindingConstants.PROPERTY_NODEID, Integer.toString(getOwnNodeId()));
     }
+
+    /**
+     * Gets the home ID associated with the controller.
+     * 
+     * @return the home ID
+     */
+    public int getHomeId() {
+        return controller.getHomeId();
+    }
 }
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -520,7 +520,7 @@ public class ZWaveThingHandler extends ConfigStatusThingHandler implements ZWave
                 ZWaveNodeSerializer nodeSerializer = new ZWaveNodeSerializer();
                 ZWaveNode node = controllerHandler.getNode(nodeId);
                 if (node != null) {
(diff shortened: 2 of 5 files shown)


mhilbush:
@cdjackson I don't think this is working when a node Thing is deleted.

When I deleted a zwave node Thing that had a node.xml file I saw this in my log (at TRACE level). After I deleted the node, the node.xml file was still there.
```
2018-07-04 09:20:59.759 [DEBUG] [rg.openhab.binding.zwave.handler.ZWaveThingHandler] - NODE 75: Handler disposed. Unregistering listener.
2018-07-04 09:20:59.761 [TRACE] [ternal.protocol.initialization.ZWaveNodeSerializer] - Initializing ZWaveNodeSerializer.
2018-07-04 09:20:59.764 [TRACE] [ternal.protocol.initialization.ZWaveNodeSerializer] - Initialized ZWaveNodeSerializer.
2018-07-04 09:20:59.764 [DEBUG] [ternal.protocol.initialization.ZWaveNodeSerializer] - NODE 75: Serializing to file /opt/openhab2/userdata/zwave/network_df9dd181__node_75.xml
2018-07-04 09:20:59.767 [DEBUG] [org.openhab.binding.zwave                         ] - ServiceEvent UNREGISTERING - {org.eclipse.smarthome.config.core.status.ConfigStatusProvider}={service.id=344, service.bundleid=185, service.scope=singleton} - org.openhab.binding.zwave
```

I see the handler being disposed and the node.xml being **serialized** as a consequence of that.

However, if `removeNode` was being called by the controller, shouldn't I expect to see another `ZWaveNodeSerializer` instance created, which would be used to invoke `nodeSerializer.deleteNode`? (Perhaps there is a need for a logging statement in `ZWaveNodeSerializer.deleteNode.` to make it clear that the serializer is deleting the node.xml?).

Also, as an aside, if it's true that `removeNode` is not being called, isn't it also the case that the node is not being removed from the controller's cache of nodes? I'm not sure of the consequence of this.

cdjackson:
> I see the handler being disposed and the node.xml being serialized as a consequence of that.

Maybe I called the wrong method (probably a cut and paste error). I'll check.

> isn't it also the case that the node is not being removed from the controller's cache of nodes

Yes, but why would we want to remove the device from the controllers cache? The node still exists - it's just the thing that is removed. I would not like to remove the device from the controller otherwise if you want to get it back you will have to reinitialise the binding.

mhilbush:
> but why would we want to remove the device from the controllers cache?

Sorry, I wasn't trying to suggest that the node be removed from the controllers cache. But, if `removeNode` is what was supposed to be called when a node Thing is deleted, that appears to be what's happening in `removeNode`. I'm not sure what mechanism you're using to detect that the framework has deleted the Thing, so I could be completely off base on what I'm saying above about `removeNode`.

PR #1016, opened 2018-09-26
Database update

cdjackson:


mhilbush:
Boo. I was using sensor_binary on the zw100 (1.7). Good thing I follow the commits, as I could've spent considerable time trying to figure out why my ZW100 stopped reporting motion... :-(

I would suggest if we're gonna clean up duplicate channels, that we do it once (anticipating a drip, drip, drip of channels I was using going away).

cdjackson:
Sorry…

I’m not sure that it can be done “once” - it requires people with the device to provide feedback. We’ve done a few recently where people with the device have “complained” about the duplication (the HeatIT thermostat is a good example) but there’s no magic solution to remove them without user input.

mhilbush:
> but there’s no magic solution to remove them without user input

I understand.

**Once** is an exaggeration, of course, but I think you know what I mean by the drip, drip, drip of this type of change.

This might be a good argument to generate an email to anyone who has set "I have this device" in the database. Probably easier than looking through all the commits...

olemr:
I have been monitoring the Z-TRM2 duplicate channels in Grafana lately and I'm not sure which one to delete because they don't always trigger/report in parallell. I have noticed that quite a few of my Fibaro devices also have double channels. Many of them have been around for quite a while and removing channels could hit many users especially  when coming to 2.4 Snapshot because then many will have to delete and re-add.

mhilbush:
> remember that nothing will actually change unless people remove the things and add them back...

True. But, you'll also get the updated channels when you include a new node. I'm sure I'm not alone in that I copy my item definitions from another node of the same type of device that's already in my network. Whoops, one device works and the other one doesn't. :scratchinghead: :confused:

I usually don't complain about things...  Sorry for venting.

cdjackson:
Sorry - not sure what to suggest. We (ie the community) remove and add channels, and change them reasonably regularly. I understand it’s a PITA, but we have 800+ devices to manage and there will be updates at times.

mhilbush:
Adding channels is no big deal (I've certainly done my share of that). I really can't speak to updating channels as I can't recall very many channels being changed.

Deleting channels is a breaking change and should be done only if absolutely necessary.

I guess we'll have to agree to disagree on this.

cdjackson:
We quite often change the channel type, and this is effectively the same thing.

I agree it should be kept to a minimum. Ideally, this would be done when the device is added, but most people do the absolute minimum when adding devices and it needs tidying later.

PR #1027, opened 2018-10-08
Fix: Actually set node to failed, proper logging for remove node

maxberger:
Issue: I am having trouble deleting a node that is no longer on my zwave network. When debugging what is happening, I found these two small obvious bugs.

Note: I still cannot confirm that this fixes my root issue (removing dead nodes), but at least now the code does what is requested (setting to failed, printing proper debug message). Unfortunately I now need to wait for my zwave network to quit down before I can investigate further.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
@@ -742,7 +742,7 @@ public abstract class ZWaveControllerHandler extends BaseBridgeHandler implement
         if (controller == null) {
             return;
         }
-        controller.requestRemoveFailedNode(nodeId);
+        controller.requestSetFailedNode(nodeId);
     }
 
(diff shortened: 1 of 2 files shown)


cdjackson:
Thanks - from a quick look, I think this is ok, but I need to check the docs. If you have a chance to test, please let me know.

cdjackson:
> where 1 would be a COMMAND_FAILED_NODE_REPLACE and 2 would be a COMMAND_FAILED_NODE_REMOVE command.

I agree with your summary of the process, but what are these two constants exactly? I don't think they come from ZWave docs, or the binding enums do they?

Using the binding enums, the process should be -:

* ReplaceFailedNode - puts the node onto the failed node list
* RemoveFailedNodeID - deletes it

The log looks ok - I don't see any reason that the controller should not return some sort of response as the command appears to be well formatted. What is the controller? Does it support these commands (I mean, it should, but worth checking)?

cdjackson:
Superceded by #1150

PR #1036, opened 2018-10-20
Implement executeRefresh on ZWaveIndicatorConverter

dfarrell-beincontrol:
Implement executeRefresh on IndicatorConverter to enable polling for indicator value.

Refresh logic is based on standard pattern found in many of the Converter classes.


cdjackson:
Have you tested this? If so, what version of the command class is it tested on? Presumably this only works on older implementations (eg V1)?

Please also sign your commit and the PR.

Thanks.

PR #1055, opened 2018-11-07
Database update

### LINKED ISSUE #1054: sensor_report channel should be write only
The ```sensor_report``` channel is only a notification channel for sending sensor reports, and should not be updated when a sensor report is received by the binding.

cdjackson:
Closes #1054 


mhilbush:
I think the change to `ZWaveMultiLevelSensorConverter.java` broke temperature processing. Shouldn't the conditional be the opposite of what was added to the code (i.e. channel == "sensor_report" versus channel != "sensor_report").

mhilbush:
Actually, I think it broke processing of all values (temp, humidity, luminance, etc) from the MULTILEVEL_SENSOR command.

cdjackson:
This is fixed - thanks.

PR #1092, opened 2018-12-20
Force associations in specific circumstances

cdjackson:
This will prevent associations being removed if the system needs them, and the controller is configured as ```master```. The PR searches to see if the database has an association configured, or if the association is the ZWave Plus ```Lifeline``` group. If so, it will set the association - thus preventing the user from removing it.

This is to avoid the system becoming unusable due to either lack of knowledge by users, or, through UI problems causing associations to be removed.

It also transpired that somewhere in the system (either in the UI, or in ESH) the arrays/lists are not being parsed correctly and the array is being presented as a string (eg ```[node_1, node_2]```). Code has now been added to handle this format!


cdjackson:
Updated to try and resolve incorrect parsing of arrays from elsewhere in the system!

PR #1100, opened 2018-12-29
Updated handling of wakeup timer

cdjackson:
This removes a check in the wakeup timer that was preventing the binding setting the device as asleep if there were messages in the queue still. The issue is if the device goes to sleep without notification, and there are messages still outstanding, then the binding will still think the device is awake, and will try to send them. This results in the device being marked dead and transactions timing out.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
@@ -95,20 +95,20 @@ public class ZWaveNode {
     @XStreamOmitField
     private boolean awake = false;
 
-    // The period to wait before telling a sleeping node to sleep again
-    private int sleepDelay = 1000;
+    // Half the period to wait before telling a sleeping node to sleep again
+    private final int sleepDelay = 1000;
 
     // Keep the NIF - just used for information and debug in the XML
     @SuppressWarnings("unused")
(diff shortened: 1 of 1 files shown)


mhilbush:
@chris I understand why you made this change, but I think I see one possible (unintended or intended?) side-effect of this change. 

During node initialization (after inclusion and/or node reinitialization), I'm seeing that many more wake-ups are required for a node to complete initialization. This seems to be a consequence of the removal of the code in the WakeupTimerTask.run method that checks for messages in the queue, as well as the reduction of the timerDelay by 50% from 5 seconds to 2.5 seconds.

I see the device being told to go back to sleep (WAKE_UP_NO_MORE_INFORMATION) repeatedly during initialization. It's especially noticeable for devices that have many command classes and config parameters.

Was it your intention to cause this change in behavior?

mhilbush:
@cdjackson Not sure if you had a chance to look at this. I realize you might still be digging out from the trip.

Hope you had a good time, and got some good pics of the wildebeest migration... If you post them online, please share the link. :-)

cdjackson:
Hi Mark,
No - I’ve not looked at this yet. I hope to do so today or tomorrow - I have some other things to finish first.

Photos are still being geotagged! I had an app running to log position, and I’m now updating the Exif data - it takes time running off the NAS as all photos were taken in RAW… Once I cut them back I’ll let you know.  And yes - it was really a great trip :)

PR #1103, opened 2018-12-31
Add all available event parameters to alarm_raw JSON data.

JesperWe:
Generalized handling of event parameter data.
This way any parameter values that are present on an event gets included in the alarm_raw JSON.
(ACCESS_CONTROL__KEYPAD_LOCK and UNLOCK will still only include the first parameter value as "code")

Signed-off-by: Jesper Weissglas <jesper@journeyman.se>


cdjackson:
@JesperWe thanks for this. Having thought about this some more, I would like to suggest a slightly different implementation.  Instead of putting your code into the ```default``` section, instead, how about putting it before or after the ```case``` statement? This way, the parameters will always be available for all events, including those that may have special treatment within the ```case``` statement.

What do you think? It should still fulfil your need, but is even more general :)

PR #1106, opened 2019-01-02
Support for disabling energy reports on Fibaro FGD212

idserda:
Currently it's not possible to completely disable energy reports on the Fibaro FGD212 module, since this parameter can not be set to 0 in the GUI. The documentation states however that this value of 0 is valid. This PR changes the minimum value for the parameter.


cdjackson:
Please update the database at https://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-database-guide. Updates directly here will be overwritten when the database is updated (every few days).

PR #1120, opened 2019-01-24
Dimmer channel changed to blinds_control channel

Fejitatete:
Tested and working.


cdjackson:
Please update the database directly. This cannot be merged as it will simply be lost when the database is updated.

Fejitatete:
Sry, I didn't realize that.

PR #1125, opened 2019-02-01
Add command class protection to support device lock

fastlorenzo:
Command class protection is needed to enable device lock.

In the manual:

> To enable Lock Mode:
• set sequence in parameter 1,
• set time and/or locking button in parameter 2 (60 seconds by default),
• set PROTECTION Command Class to Local Protection by Sequence (done automatically by home Center controller)


cdjackson:
This will require a database update. If we merge this here, it will be removed at the next database update (ie within a couple of days).

fastlorenzo:
How can I help with this?

PR #1129, opened 2019-02-10
Removed extraneous byte in Network Security Key example

vzorglub:


cdjackson:
Isn't this the same as the other one? Why not fix all instances of this error - otherwise you are making this worse. At least at the moment it's reasonably obvious that it's an error - by changing just one example to be correct, and leaving the other 3 incorrect, this is now really confusing.

Please can you update this as requested previously.

vzorglub:
My bad, I only noticed that the last one was bad.
Let's try again...

vzorglub:
Third time lucky #1130

PR #1134, opened 2019-02-15
Update wakeup timer during initialisation

cdjackson:
Use a longer wakeup timer during initialisation even if there are currently no messages in the queue.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
@@ -135,8 +135,8 @@ public class ZWaveNode {
     /**
      * Constructor. Creates a new instance of the ZWaveNode class.
      *
-     * @param homeId the home ID to use.
-     * @param nodeId the node ID to use.
+     * @param homeId     the home ID to use.
+     * @param nodeId     the node ID to use.
      * @param controller the wave controller instance
      */
     public ZWaveNode(int homeId, int nodeId, ZWaveController controller) {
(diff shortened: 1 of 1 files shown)


mhilbush:
Does this change improve the secure inclusion issue for battery devices? Seems like it also will improve the situation with initialization of battery devices requiring many wake ups.

cdjackson:
Yes, that is the plan.

And thanks for looking at it Mark - I thought you would, which is why I’ve not yet merged it :)

cdjackson:
Of course, I should have added that while it's the plan that it helps batter device initialisation, let's see how it performs ;)

mhilbush:
I'll install the next build, then put it through its paces by reinitializing a few devices that I was having a hard time initializing. The removal of the check for an empty queue when initialization not complete may have the biggest effect of forcing the 5 second timer. I may not be able to do this until tomorrow.

cdjackson:
Thanks Mark,
I gave this a basic test, but it’s not always possible to ensure that there is nothing in the queue when a device wakes during initialisation ;)

Any issues, just yell :)

mhilbush:
My initial testing of this shows some improvement. Initialization of a Fibaro FGMS001 had required from 2 to 3 dozen wake-ups in order for the device to complete initialization. After installing the updated binding, it took 11 wake-ups for it to complete.

Of course, the number of wake-ups for any given device will vary, based on the number of config parameters, association groups, and command classes supported by the device. The FGMS001 will be on the high side because it has a lot of config parameters and command classes.

IMO, it still would be good to try to improve on this. Is there a downside to setting the timer to something greater than 5 seconds during initialization? Is 8-10 seconds out of the question?

Here's a log. The FGMS001 is node 112.
https://drive.google.com/open?id=1BnwSgYC5rLkGPjF_rnxlCK1_8dlRHWan

As an aside, the Wallmote swipe transactions (config reports for param9 and param10) also are in this log. The Wallmote is node 72.

PR #1136, opened 2019-02-17
Fix COMMAND_CLASS_SWITCH_ALL to correctly handle only on/off setting

MFornander:
The listen to only ON or only OFF was swapped as comppared to the specs at:
https://www.silabs.com/documents/login/miscellaneous/SDS13781-Z-Wave-Application-Command-Class-Specification.pdf  page 30.
Also verified that setting to "Only ON" on OpenHAB shows up as "Only OFF" in Z-Wave PC Controller 5 utility.

Note that this is dry coded since I haven't leaned how to build and install a custom AddOn in OpenHAB yet.  Would love a link or hint where to start so I can be more productive.


cdjackson:
Thanks for the correction. I suspect that there are now failing tests so please look at these, and also please sign the commit and PR (see the contributing guidelines).

There is a description of how to install the IDE on the web - just go to the openhab documentation and follow the developper guide pages.

MFornander:
Nice CI setup and tests.  I'll read through the developer guidelines and get it building locally before I commit the fix.

PR #1142, opened 2019-02-26
Remove deprecated ExtendDiscoveryService API

hanneshofmann:
The `ExtendDiscoveryService` API has been marked as deprecated and returns always null, thus it can be safely removed. This PR removes the usage of the API without changing the behaviour of the `ZWaveDiscoveryService`.

Signed-off-by: Hannes Hofmann <hannes.hofmann@telekom.de>


cdjackson:
Thanks Hannes.

PR #1147, opened 2019-03-03
Database update

cdjackson:


apica-alfred:
I think there's an extra change not related with the database update (src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveMultiLevelSwitchConverter.java).

cdjackson:
Yes - what is the issue your reporting - or are you just making the point?

apica-alfred:
Just making the point just in case you added some extra changes without noticing it. The typical "Database update" commits don't make any other changes.

PR #1150, opened 2019-03-03
Fix incorrect call to RemoveFailedNode

cdjackson:


src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
```diff
@@ -1001,10 +1001,10 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
                 case "action":
                     if ("failed".equals(cfg[1]) && valueObject instanceof Boolean && ((Boolean) valueObject) == true) {
                         controllerHandler.replaceFailedNode(nodeId);
+                        controllerHandler.checkNodeFailed(nodeId);
```
mhilbush: This looks right, but the terminology is a bit confusing...

The method name **replaceFailedNode** is confusing, because it sends the SetFailedNode request. I suspect this ties back to the zwave terminology, but it's confusing nonetheless.

cdjackson: I do agree, but as you rightly assume, I tried to stick to the ZWave terminology (and is probably why I cocked it up in the first place).

PR #1183, opened 2019-06-05
Fix for PROTECTION_SET V2 (#1182)

### LINKED ISSUE #1182: PROTECTION_SET V2 does not work when setting local protection only
* Binding version: latest snapshot

When trying to set local protection type, `setValueMessage` is invoked with the second parameter equal to `null`. (https://github.com/openhab/org.openhab.binding.zwave/blob/master/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveProtectionConverter.java#L85)

Due to `rfMode == null` check here: https://github.com/openhab/org.openhab.binding.zwave/blob/master/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveProtectionCommandClass.java#L192

Only one byte is sent (like in version 1) and the message is rejected on devices supporting version 2 of this command (example of such device is FGBS-222).

message sent to the node:

NODE 20: Sending REQUEST Message = 01 0E 00 13 14 07 60 0D 01 05 **75 01 02** 25 92 59

rmichalak:
Fixes #1182

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveProtectionCommandClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveProtectionCommandClass.java
@@ -73,6 +73,9 @@ public class ZWaveProtectionCommandClass extends ZWaveCommandClass
     @XStreamOmitField
     private LocalProtectionType currentLocalMode;
 
+    @XStreamOmitField
+    private RfProtectionType currentRfMode;
+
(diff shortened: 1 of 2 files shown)


rmichalak:
Ok, the build failed spectacularly, but I don't see my commit causing it.. ;)

cdjackson:
No - it's not your PR - there are a lot of problems with the build these days :(  I'll take a look at this tomorrow - thanks.

PR #1185, opened 2019-06-12
Don't clear association groups if only controller has been added

jimzrt:
related to [Association Groups lost after saving or restarting openHAB [debug log attached]](https://community.openhab.org/t/association-groups-lost-after-saving-or-restarting-openhab-debug-log-attached/76155/11)

all new group members are added up to `totalMembers` **if** they are not the controller and if `totalMembers == 0`, then all associations in this group are cleared.
Since added controller in groups don't count to totalMembers, the group is always cleared.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -734,7 +734,6 @@ public class ZWaveThingHandler extends ConfigStatusThingHandler implements ZWave
                     logger.debug("NODE {}: Current members before update {}", nodeId, currentMembers);
 
                     ZWaveAssociationGroup newMembers = new ZWaveAssociationGroup(groupIndex);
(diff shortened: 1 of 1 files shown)


src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
```diff
@@ -797,7 +794,8 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
                         }
                         logger.debug("NODE {}: Members after controller update {}", nodeId, newMembers);
                     }
-
+                    
+                    int totalMembers = currentMembers.getAssociationCnt() + newMembers.getAssociationCnt();
                     // If there are no known associations in the group, then let's clear the group completely
                     // This ensures we don't end up with strange ghost associations
                     if (totalMembers == 0) {
```
cdjackson: I believe that this is incorrect. This should simply be -:

```
if(newMembers.getAssociationCnt() > 0)
```

cdjackson:
Thanks.

PR #1186, opened 2019-06-14
Change registration of discovery to handler factory

jimzrt:
I'm trying to get to know the code base and stumbled upon a deprecated access to bundleContext in the BaseThingHandler.
I saw the already fixed issue for the ZigBeeBinding ([#31](https://github.com/openhab/org.openhab.binding.zigbee/issues/31)) and the corresponding pull. [#34](https://github.com/openhab/org.openhab.binding.zigbee/pull/34)

So I tried to do the same here, though I'm not quite sure if I got everything right. (Particularly adding a DiscoveryService for the ZWaveSerialHandler in ZWaveHandlerFactory)

I'm still reading and learning, so if I'm way off track with this pull request, please say so.

Hope, I can contribute somewhat.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/discovery/ZWaveDiscoveryService.java
+++ b/src/main/java/org/openhab/binding/zwave/discovery/ZWaveDiscoveryService.java
@@ -28,7 +28,12 @@ import org.openhab.binding.zwave.ZWaveBindingConstants;
 import org.openhab.binding.zwave.handler.ZWaveControllerHandler;
 import org.openhab.binding.zwave.internal.ZWaveConfigProvider;
 import org.openhab.binding.zwave.internal.ZWaveProduct;
+import org.openhab.binding.zwave.internal.protocol.ZWaveEventListener;
 import org.openhab.binding.zwave.internal.protocol.ZWaveNode;
+import org.openhab.binding.zwave.internal.protocol.event.ZWaveEvent;
+import org.openhab.binding.zwave.internal.protocol.event.ZWaveInclusionEvent;
+import org.openhab.binding.zwave.internal.protocol.event.ZWaveInitializationStateEvent;
+import org.openhab.binding.zwave.internal.protocol.serialmessage.ZWaveInclusionState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -38,7 +43,7 @@ import org.slf4j.LoggerFactory;
  * @author Chris Jackson - Initial contribution
  *
  */
-public class ZWaveDiscoveryService extends AbstractDiscoveryService {
+public class ZWaveDiscoveryService extends AbstractDiscoveryService implements ZWaveEventListener {
     private final Logger logger = LoggerFactory.getLogger(ZWaveDiscoveryService.class);
 
     private final String ZWAVE_NODE_LABEL = "Z-Wave Node %03d";
@@ -54,11 +59,13 @@ public class ZWaveDiscoveryService extends AbstractDiscoveryService {
 
     public void activate() {
         logger.debug("ZWave discovery: Active {}", controllerHandler.getThing().getUID());
+        controllerHandler.addEventListener(this);
     }
 
     @Override
     public void deactivate() {
         logger.debug("ZWave discovery: Deactivate {}", controllerHandler.getThing().getUID());
+        controllerHandler.removeEventListener(this);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


cdjackson:
Please rebase this and also run the project formatter over it and I will take a look.

jimzrt:
alright, should be correct now

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -562,14 +554,13 @@ public void ZWaveIncomingEvent(ZWaveEvent event) {
                     break;
                 case IncludeFail:
                     eventKey = ZWaveBindingConstants.EVENT_INCLUSION_FAILED;
-                    eventState = BindingEventType.WARNING;
```
cdjackson: Why change this? I'd prefer to leave it as is.

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -701,20 +686,29 @@ public void sendData(ZWaveCommandClassTransactionPayload transaction) {
         controller.sendData(transaction);
     }
 
-    public boolean addEventListener(ZWaveThingHandler zWaveThingHandler) {
+    public boolean addEventListener(ZWaveEventListener listener) {
+
```
cdjackson: Please remove the blank line.

src/main/java/org/openhab/binding/zwave/discovery/ZWaveDiscoveryService.java
```diff
@@ -228,4 +235,15 @@ public void deviceAdded(ZWaveNode node) {
...
+
+    @Override
+    public void ZWaveIncomingEvent(ZWaveEvent event) {
+        if (event instanceof ZWaveInclusionEvent) {
+            ZWaveInclusionEvent incEvent = (ZWaveInclusionEvent) event;
+            if (incEvent.getEvent() == ZWaveInclusionState.IncludeDone && incEvent.getNodeId() != 0) {
+                deviceDiscovered(event.getNodeId());
+            }
+        }
+
```
cdjackson: Please remove the blank line.

src/main/java/org/openhab/binding/zwave/discovery/ZWaveDiscoveryService.java
```diff
@@ -54,11 +58,14 @@ public ZWaveDiscoveryService(ZWaveControllerHandler coordinatorHandler, int sear
 
     public void activate() {
         logger.debug("ZWave discovery: Active {}", controllerHandler.getThing().getUID());
+        controllerHandler.addEventListener(this);
+
```
cdjackson: Please remove the blank line.

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -653,17 +644,11 @@ protected void incomingMessage(SerialMessage serialMessage) {
 
     @Override
     public void deviceDiscovered(int nodeId) {
```
cdjackson: Why not just delete this method?

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -653,17 +644,11 @@ protected void incomingMessage(SerialMessage serialMessage) {
...
-        if (discoveryService == null) {
-            return;
-        }
-        // discoveryService.deviceDiscovered(nodeId);
+        // Nothing to do here
     }
 
     public void deviceAdded(ZWaveNode node) {
```
cdjackson: Why not just delete this method?

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
jimzrt: I thought it would be more consistent because `SecureIncludeFailed` uses `BindingEventType.ERROR` as well. 
I will revert it - it doesn't relate to the pull request, so it shouldn't be done here

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
jimzrt: right, `ZWaveControllerHandler `is an abstract class, so it can be deleted here, but then I need to add it to `ZWaveSerialHandler`, because it extends `ZWaveControllerHandler `.
Is it better to do it that way?

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
jimzrt: I deleted it and there was a special case that should be handled by the event handler in `ZWaveDiscoveryService` - I added this case

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
cdjackson: Well I think you need to understand why there is this method as you've stopped it doing anything. If I remember correctly it is part of the IoHandler interface so if it's not used, it should be removed - if it is used, it should be used?

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
cdjackson: Ok - sounds fine.

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
jimzrt: Hm, it doesn't do anything if I see it correctly.
The call to the discoveryService in `ZWaveControllerHandler` was commented out before and there is no other implementation of this method.
DeviceDiscovery is handled via events in initial and later stage, so there is no need, as far as I can see it.

So I would propose to remove the method in the IoHandler interface.
I did that and all tests run successfully.

Should I add this change to the pull?

cdjackson: Thanks. Yes, this is what I was aiming for :)

jimzrt:
I think all your suggestions are taken into account now!
please take a look when you get a chance

PR #279, opened 2016-12-25
Dimmer Restore previous value

marcelrv:
addresses #267

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

PR #1188, opened 2019-06-17
update handling of association

jimzrt:
fixes [#279](https://github.com/openhab/org.openhab.ui.habmin/issues/279)
there have been multiple reports that there are issues when displaying association groups in habmin (e.g [here](https://community.openhab.org/t/oh2-z-wave-refactoring-and-testing-and-security/21653/3003))

it was assumed that the used widget is at fault, but I think it is a consequence of pull [#953](https://github.com/openhab/org.openhab.binding.zwave/pull/953)

when synchronizing the group association configuration, the node id of the controller has to be replaced with the proxy constant as well, because the configuration in openhab expects it.
The field in habmin is empty, because there is no option for the reported node id as group member from the device if it is the controller.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -724,7 +724,7 @@ public class ZWaveThingHandler extends ConfigStatusThingHandler implements ZWave
                             paramValues.add(strParam);
                         }
                     }
-                    logger.debug("NODE {}: Association {} consolidated to {}", nodeId,groupIndex, paramValues);
+                    logger.debug("NODE {}: Association {} consolidated to {}", nodeId, groupIndex, paramValues);
 
                     ZWaveAssociationGroup currentMembers = node.getAssociationGroup(groupIndex);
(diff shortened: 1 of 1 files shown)


jimzrt:
correction: the referenced pull request was made after the reported issue, so this just fixes the associatio

cdjackson:
It's unclear to me how this can impact HABmin. The HABmin widget definately has an issue - are you now saying that this fixes that without even changing anything in HABmin? Seems a bit weird to me given what this PR does - can you explain how it resolves this?

Thanks.

jimzrt:
as far as I understood, the `ZwaveConfigProvider` is responsible for possible options in a select field in the UI.
For the association, the controller constant is defined on top and then all other nodes except the controller.

When setting the configuration, all is good and the field is populated with the 'controller' constant.
I just noticed that after a restart of openhab, the config field is populated not with the "controller" constant but the actual node id. (all verified via rest api)

So without deep diving into the widget code I just replaced the node id with the controller again when reading the configuration back, because I thought that the widget cannot preselect a value that is not present in all the options.

In my case that did the trick, but maybe this is just my particular case and not the general one...

Issue #43, opened 2016-08-21
Aeotec Minimote failing to be recognized


mrzacharycook:
When attempting to add a DSA03202B-ZWUS Aeotec Minimote to Openhab2 via autodiscovery, the following is logged:

2016-08-21 05:02:53.081 [ERROR] [ve.internal.protocol.ZWaveController] - Neither inclusion nor exclusion was active!
2016-08-21 05:02:53.186 [WARN ] [wave.discovery.ZWaveDiscoveryService] - NODE 2: Device could not be resolved to a thingType! 7FFFFFFF:7FFFFFFF:7FFFFFFF::0.0
2016-08-21 05:02:53.411 [INFO ] [smarthome.event.InboxAddedEvent     ] - Discovery Result with UID 'zwave:device:36e36005:node2' has been added.
2016-08-21 05:02:53.415 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'zwave:device:36e36005:node2' to inbox.

Attributes from the Thing configuration page are:

zwave_class_basic	CONTROLLER
zwave_class_generic	REMOTE_CONTROLLER
zwave_deviceid	2147483647
zwave_frequent	false
zwave_nodeid	2
zwave_version	0.0
zwave_listening	false
zwave_routing	false
zwave_beaming	true
zwave_class_specific	PORTABLE_REMOTE_CONTROLLER
zwave_manufacturer	2147483647
zwave_devicetype	2147483647

Any ideas why it wouldn't be picking up the deviceid correctly?

cdjackson:
Please read the readme for this binding. I know this isn't currently compiled into the docs due to the separate repository but it should be soon. 

Cheers
Chris

cdjackson:
Just to be clear - this is the main bit I was referring to -:

Unknown Devices

If the device is listed as Unknown, then the device has not been fully discovered by the binding. There are a few possible reasons for this -:

The device is not in the database. If the device attributes show that this device has a valid manufacturer ID, device ID and type, then this is likely the case. Even if the device is in the database, some manufacturers use multiple sets of references for different regions or versions, and your device references may not be in the database. In either case, the database must be updated and you should raise an issue to get this addressed.
[...]

PR #1193, opened 2019-07-01
revise association handling

jimzrt:
Hi @cdjackson ,

I know this is the third pull request regarding association handling, but just like last time it fixes my particular problems and maybe you deem it a general problem / fix, so I just want to put it out there.

I tried to find out what is really causing UI problems regarding group associations and after the last pull request, habmin is working correctly for me whereas paperUI is not. (short discussion [#43](https://github.com/openhab/openhab-webui/issues/43))

paperUI displays associations depending on the device's configuration and expects this configuration in a specific format: 
if the group can have multiple members, it expects a list, and if not, it expects just the member.

The group configuration in the binding is set irregardless of the device being able to have multiple members - it returns as a list if there are multiple members and it returns the first member if it is currently the only member. When there is no member currently set, it returns an empty string.

I changed it so that the device's configuration is taken into account in the ZWaceThingHandler and it sets and return its group association members depending on the device being able to have multiple members in that particular group.
I.e. if the device can have multiple members, a list is always returned (even if there are no members).

I think this behaviour is more intuitive and what I would expect when asking for or setting group members.
Additionally, paperUI and habmin are working without any problems when setting and displaying group associations now.

I added a few fixes like preventing to add nodes as group members if the node is not know by the controller or adding multiple members if the device can only have one member.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -31,6 +31,8 @@ import java.util.TimeZone;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
+import org.eclipse.smarthome.config.core.ConfigDescription;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.config.core.status.ConfigStatusMessage;
 import org.eclipse.smarthome.config.core.validation.ConfigValidationException;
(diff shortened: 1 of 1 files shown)


cdjackson:
I would probably prefer not to buffer the configuration locally in the thing handler as it could get out of sync with the real configuration. I don't think there's a need to store it as a field in the class is there?

I'd also question if it's the best approach in general or if we shouldn't just use the ```maxNodes``` field in the ```AssociationGroup``` class which is already available here. This would remove the dependancy on the configuration which I think it nicer, but it has the drawback that if the database is defined differently than what the device reports, there will be a problem - my guess if this happens is there will be a problem anyway though!

cdjackson:
Oh - and I meant to add - thanks for looking at this :)

lsiepel:
Does this PR also fix this issue? https://github.com/openhab/org.openhab.binding.zwave/issues/1205
I'm happy to test this, but someone has to provide a jar file before i can do so.

lsiepel:
@jimzrt any progress on this? I'm happy to test as i still have problems with the associations.

PR #1204, opened 2019-07-18
Update README.md

robmachab:
Please review. 

While I am a native UK english speaker, I suffer from Poo'ism, an affliction of the bear type. Things can get muddled and also possibly are not great for international english speakers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Sets the controller as a Static Update Controller within the network
 
 #### Heal Time [heal_time]
 
-Sets the nightly heal time (in hours).
+Sets the nightly heal time (in hours). This is best set to disabled unless your network has routing issues. Be aware that a full network heal may cause more issues than you already have so it may be better to tackle the nodes that have issues. The setting can be found under the controller node in things in HABmin. To heal an individaul node there is a command under tools advanced settings in HABmin. It should not be necessary to use heal in a stable network and particularly in a Z-Wave+ network. see Mesh Heal later on this page for more details.
 
 
 #### Inclusion Mode [inclusion_mode]
@@ -283,13 +283,13 @@ A special type of battery device is a *FLiRS* device. These devices do not sleep
 
 ### Polling
 
-The binding supports periodic polling. This has two purposes - firstly to ensure that a device is still responding, and secondly to update the bindings representation of the state of the device.  Where possible *associations* should be used to update the device state - this will keep network traffic to a minimum, which will improve the network latency, and it will be faster since *associations* send updates to the controller immediately where polling will always be noticeably slower.
+The binding supports periodic polling. This has two purposes - firstly to ensure that a device is still responding, and secondly to update the bindings representation of the state of the device.  Where possible *associations* should be used to update the device state and polling set to days or weeks - this will keep network traffic to a minimum, which will improve the network latency, and it will be faster since *associations* send updates to the controller immediately where polling will always be noticeably slower.
 
 If a device fails to respond to a poll, then it will be marked as DEAD and shown as offline. For battery devices, if they do no provide a wakeup within a period of twice the wakeup period, then they will also be considered dead and taken offline.
 
-Keep the polling at a slow rate unless your device doesn't support *associations*. This will reduce network traffic, reduce the chance of timeouts and retries, and therefore improve the overall performance of the network.
+Keep the polling at a slow rate many hours, days or weeks unless your device doesn't support *associations*. This will reduce network traffic, reduce the chance of timeouts and retries, and therefore improve the overall performance of the network.
 
-The binding can perform a poll of the device shortly after sending a command to make sure that the command was implemented, and the binding has the correct view of the devices state. This is called "Command Poll Period" and may need adjustment for some devices that may update their state slowly (e.g. dimmers that have a slow transition). This is defined in milliseconds, and can be set to 0 to disable this polling.
+The binding can perform a poll of the device shortly after sending a command to make sure that the command was implemented, and the binding has the correct view of the devices state. This is called "Command Poll Period" and may need adjustment for some devices that may update their state slowly (e.g. dimmers that have a slow transition). This is defined in milliseconds, and can be set to 0 to disable this polling. Again polling for most devices is not required and should only be set if there is a noticable issue with reporting from the device.
 
 
 ### Binding Maintenance Functions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -283,13 +283,13 @@ A special type of battery device is a *FLiRS* device. These devices do not sleep
 
 ### Polling
 
-The binding supports periodic polling. This has two purposes - firstly to ensure that a device is still responding, and secondly to update the bindings representation of the state of the device.  Where possible *associations* should be used to update the device state - this will keep network traffic to a minimum, which will improve the network latency, and it will be faster since *associations* send updates to the controller immediately where polling will always be noticeably slower.
+The binding supports periodic polling. This has two purposes - firstly to ensure that a device is still responding, and secondly to update the bindings representation of the state of the device.  Where possible *associations* should be used to update the device state and polling set to days or weeks - this will keep network traffic to a minimum, which will improve the network latency, and it will be faster since *associations* send updates to the controller immediately where polling will always be noticeably slower.
```
cdjackson: I'm not sure I would agree here. Polling is useful to detect if a device is DEAD - normally you want to detect that more often than every week or two. The reality is that polling even with a rate of a few hours has an extremely low impact - eg with 200 nodes on the network, and a poll rate of 4 hours, we would have network usage by polling of about 1 frame every 15 to 20 seconds, and an overall loading of around 0.02% - not so bad ;)

README.md
```diff
@@ -283,13 +283,13 @@ A special type of battery device is a *FLiRS* device. These devices do not sleep
...
 
 If a device fails to respond to a poll, then it will be marked as DEAD and shown as offline. For battery devices, if they do no provide a wakeup within a period of twice the wakeup period, then they will also be considered dead and taken offline.
 
-Keep the polling at a slow rate unless your device doesn't support *associations*. This will reduce network traffic, reduce the chance of timeouts and retries, and therefore improve the overall performance of the network.
+Keep the polling at a slow rate many hours, days or weeks unless your device doesn't support *associations*. This will reduce network traffic, reduce the chance of timeouts and retries, and therefore improve the overall performance of the network.
```
cdjackson: Same comment - why days or weeks? It does not make sense I think?

README.md
```diff
@@ -283,13 +283,13 @@ A special type of battery device is a *FLiRS* device. These devices do not sleep
...
 
-The binding can perform a poll of the device shortly after sending a command to make sure that the command was implemented, and the binding has the correct view of the devices state. This is called "Command Poll Period" and may need adjustment for some devices that may update their state slowly (e.g. dimmers that have a slow transition). This is defined in milliseconds, and can be set to 0 to disable this polling.
+The binding can perform a poll of the device shortly after sending a command to make sure that the command was implemented, and the binding has the correct view of the devices state. This is called "Command Poll Period" and may need adjustment for some devices that may update their state slowly (e.g. dimmers that have a slow transition). This is defined in milliseconds, and can be set to 0 to disable this polling. Again polling for most devices is not required and should only be set if there is a noticable issue with reporting from the device.
```
cdjackson: Unfortunately this is incorrect. Polling is needed here to read back the state of the device to confirm that the command was implemented. We could just assume it is ok of course, but there's no other way to establish this without polling.

README.md
robmachab: Yes agree with the dead device but this should be very rare and certainly formy spare room every week check is fine. As I posted this is the reason I will leave it on for devices I rarely use.

For some i use more often I keep it at a few hours. For those I use very regularly I used to turn it off on HC2 as it was just not needed.

On the utilisation I am not convinced that your calculation is correct for all situations though as the silicon improves it may get closer to the theory. Also many devices when polled will return a few more than 1 frame but agree totally that every 4 hours should not harm a good network. I suppose it is an optimist vs  pessimist discussion we are having. It really needs to be fit for your needs but some people are a bit daft and think a kWh report every 0.1 s from every device even it not drawing power is useful. If we could get everybody to start with very little then adjust the ones that need more regular it would be a safe approach.

cdjackson: You are free to do what you like and if you want to set the polling to 10 days, then this is up to you. However here you are making recommendations for the documentation, and we should be more generic and you should therefore consider what you are recommending for the masses, and not just your personal preference.

Node DO drop off the network - it is not so uncommon, and it may happen due to variable propagation during the day.

The calculation is simple - 200 devices with an average of 4 channels is 800 requests. If we assume these are spread over the 4 hours (the binding tries to avoid them all being together!), then that is 800 requests every 14400 seconds, or 1 request every 18 seconds. At 100kb/s, this is a loading that is around the 0.02%. Maybe this isn't 100% accurate, and it assumes the requests get spread over time which again isn't 100% accurate, but it's in this order of magnitude, and it shows the loading is (very!) low.

We are NOT talking about polling every 0.1s - I don't know why you keep raising this.

robmachab: 4 hours would be a good start. As I say optimist vs pessimist but we appear to be the reverse on different issues. I suppose we gravitate to the issue that has bitten us most over the years. I like the sound of a stable network with 200 nodes with lots of reporting and polling every 4 hours. Have you ever managed that? If you have I know a few HC2 users that will be over in a shot.

As to the talk about reporting, Sorry as I say I am often not clear in coms. I think you are extremely precise. Some days i wish I was.

What I was talking about was combining polling with reporting as the paragraph addresses both.
 
The two basic things new users fail to understand as we see in the discussion on this board and over at Fibaro.

Far too many new users I have dealt with over the years configure polling to every minute and reporting changes every 10s for every device as they configure everything to report on a small change. They get to 30 devices with 10 sleeping nodes with an amount of routing. Then start seeing slow updates so make it even worse by configuring some even more regular polling and reporting as that seems to be the issue.

Your original text should be enough to stop them but possibly recommending 4 hours as a starter is correct and would make it better so can we agree on that?.



If we limit the change to that for now.

It would be interesting to have a further discussion on bandwidth as I still feel it saturates well below the theoretical level when I have tested. I have never put my finger on why as it seems that it should be possible to send a lot of frames as you describe. Possibly other things are going on that are not shown with a zniffer.  I know if I make users think it is less than it is. it takes a lot less time for the issue to go away That is a good result if a bit disingenuous and lacking your precision.

Anyway thanks for the invite and the challenges.  Will make the one small adjustment for now and just suggest 4hrs. It has to be better than the settings some users are trying so certainly will make things a lot better for them if they bother to read.

README.md
majennah: 10 hrs ago

README.md
cdjackson: Please remove the days to weeks part as it's not possible to set this above 10 days.

README.md
```diff
@@ -56,7 +56,7 @@ Sets the controller as a Static Update Controller within the network
 
 #### Heal Time [heal_time]
 
-Sets the nightly heal time (in hours).
+Sets the nightly heal time (in hours). This is best set to disabled unless your network has routing issues. Be aware that a full network heal may cause more issues than you already have so it may be better to tackle the nodes that have issues. The setting can be found under the controller node in things in HABmin. To heal an individaul node there is a command under tools advanced settings in HABmin. It should not be necessary to use heal in a stable network and particularly in a Z-Wave+ network. see Mesh Heal later on this page for more details.
```
cdjackson: Please can you remove this comment.

README.md
```diff
@@ -343,6 +343,23 @@ If the device is listed as *Unknown*, then the device has not been fully discove
...
+When configuring individual devices and writing automation scripts the following may help avoid a common pitfall of transmitting too much data in a short time.
+
+The controller stick can only be sent one message at a time from the binding to be transmitted to the network.  Any command sent when another message is being transmitted is placed in a queue in the binding.
+
+A message can take anything from just over 50 ms to many 100s of ms to be transmitted and a response to come back from the node across the network. There is therefore a finite number of messages that can be sent and received in a second while maintaining a sensible depth of queue. If too many messages are sent for too long a period then there will be noticeable pause when manually controlling devices.
+
+If a queue has many messages from an automation script, a user should not be surprised if there is a delay when interactively controlling nodes while the script runs. If the automation will run during a period when manual control may also be used, a pause between each command sent by the automation will improve user experience as it makes it less likely that the interactive command is queued behind many commands.
```
cdjackson: This comment about adding the delay isn't really true. The binding will not send messages until the previous response is received, so there is no need for a delay. Additionally, the binding will prioritise messages, so adding a delay won't necessarily move things around as the user expects as order is not maintained. I would suggest to remove this part as it's likely to cause more confusion than assistance.

README.md
```diff
@@ -343,6 +343,23 @@ If the device is listed as *Unknown*, then the device has not been fully discove
...
+
+In general only configure devices to send reports needed at the lowest frequency they are useful and disable any reports of values not needed. Most modern device allow configuration of the time interval or amount of change in a value to trigger a report. Some even allow the specification of the minimum time between consecutive reports. It is better to be conservative with these values as a network is built to avoid a lot of tuning and configuration later.
+
+When a node fails in some way, the bandwidth becomes strangled even further due to the long time a stick tries to talk to the failed node. The stick will try several times to talk to the failed node each with a timeout before returning a failure to the Z-wave binding. While this is happening no other message can be transmitted. If the device is in a long term failed sate this long pause will be repeated as the OpenHAB binding also issues a retries. If there is a lot of regular traffic when this happens very undesirable results are often observed with many devices appearing to have issues despite the issue possibly being caused by a single device.
```
cdjackson: Can you just change the word "stick" to controller.

cdjackson:
Closing this as there's been no update for 6 months.

PR #1207, opened 2019-07-22
ThingHandlerFactory: start on demand

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@free.fr>

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/ZWaveHandlerFactory.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/ZWaveHandlerFactory.java
@@ -45,7 +45,7 @@ import org.slf4j.LoggerFactory;
  *
  * @author Chris Jackson - Initial contribution
  */
-@Component(immediate = true, service = { ThingHandlerFactory.class })
+@Component(service = ThingHandlerFactory.class)
 public class ZWaveHandlerFactory extends BaseThingHandlerFactory {
     private Logger logger = LoggerFactory.getLogger(BaseThingHandlerFactory.class);
 


lolodomo:
It seems to have fixed the problem I got at OH startup.
But of course, using now a new jar file in folder addons could have changed the order of started bundles.
But by the way, this is better to not have this "immediate = true".

cdjackson:
Shouldn’t we have a common way to implement this in openHAB? If I look in the openhab addons most binding seem to have immediate=true so it seems strange to change this.

lolodomo:
In OH2, most of them have not "immediate = true" and i just created a PR to suppress "immediate = true" from the 15 bindings for which it was still there.
See the discussion in OH2 and the multiple previous PR that already tried to suppress "immediate = true".

cdjackson:
Ok, thanks. Please link the other PR and once it’s merged I’ll merge this.

lolodomo:
Here is the opned OH2 PR: https://github.com/openhab/openhab2-addons/pull/5855
And the discussion: https://github.com/openhab/openhab2-addons/issues/5852

openhab-5iver:
@lolodomo, should the same be done for [zigbee](https://github.com/openhab/org.openhab.binding.zigbee/search?q=thinghandlerfactory&unscoped_q=thinghandlerfactory) too?

lolodomo:
Yes, of course.

lolodomo:
The OH2 PR is now merged. 
Please do the same for this one.

cdjackson:
Thanks.

lolodomo:
Can you update the zigbee binding if necessary?

PR #1209, opened 2019-07-25
updateStatus: Fix error messages containing parameter

lolodomo:
Fixes issue #1014

Also add status CONFIGURATION_ERROR.

Also reduce log level in the thing handlers

Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #980, opened 2018-08-22
Implement reconnection to ZWave serial stick (#979)

dottedmag:
Keep probing the port if the stick is not ready on startup or an error has been
received from the serial port.

PR #1210, opened 2019-07-26
Several improvements on serial port stability

### LINKED ISSUE #979: Implement reconnect if device is not ready or disconnected
It would be nice if Z-Wave binding was reconnecting to a device if it is not ready on startup or was disconnected and later reconnected (like Aeotec USB dongle).

I have an implementation and will submit a PR soon.

Hilbrand:
- Added watchdog to handle serial port initialization. This makes serial port more stable as it can handle when the serial port goes down or isn't available yet.
- Improved localized status messages by adding portId as parameter argument.
- In ZwaveControllerHandler the listeners added externally are not added directly to the controller anymore. This improves stability because it removed the dependency of the availability of the controller.
- Used cron scheduler for healing task

Closes #979, supersedes #980
 
**NOTE:** In #980 comments it was mentioned the whole network management might be affected by a change like this and might need some rework. I didn't fully tested it, so that part might still be blocking for this pr. But I think this pr might be a starting point on improving stability.
**NOTE 2:** It also mentioned to create a pr against development, but that branch seems out-of-date. So I created it against master.
**NOTE 3:** I just noticed this pr covers some changes suggested in #1209 (I probably need to update some logging here :slightly_smiling_face:)

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
@@ -15,7 +15,6 @@ package org.openhab.binding.zwave.handler;
 import static org.openhab.binding.zwave.ZWaveBindingConstants.*;
 
 import java.math.BigDecimal;
-import java.util.Calendar;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


cdjackson:
Thanks. I've not looked though this in detail, but I'm definitely concerned that this will not reinitialise the network. From what I see, the restart is handled solely in the comms handler, and does not account for the system level impacts? I don't see the dispose method closing down all the nodes so that they can be restarted later once the stick is reinitialised.

I would like to understand how this is handled or we will either end up with the controller not being initialised properly, and/or the nodes being duplicated, missing, or memory leaks?

I believe that in order to do this properly, we should shutdown the network, and reinitialise as if the binding had been restarted since you can't know what has happened while the dongle is down. For example, I see people promoting this sort of thing as a workaround to allow the stick to be removed from the PC to include new devices - therefore when it is plugged back in, the network will have changed and we need to handle this sort of thing.

cdjackson:
@Hilbrand I don't think that conceptually this is the right approach. The reinitialisation should not be handled in the serial port driver - it must be handled in the controller serial handler. Maybe the serial port driver may need to detect something wrong with the serial port, but all reconfiguration must be handled by closing down all nodes, closing the port, then reinitialising as if starting the binding again.

Fundamentally this is also how we handle this in ZigBee.

So, I think your comment that this is a good starting point is unfortunately probably incorrect since I think it needs a rethink and rearchitecting of the controller handler.

How do you want to handle this?

cdjackson:
Closing this as it's not the right approach.

cdjackson:
#1213

PR #1222, opened 2019-08-14
Add serial port reconnection logic (Closes #1213)

### LINKED ISSUE #1213: Add serial port reconnection logic
The binding would be more reliable and easier to use when it has its own serial port reconnection logic (see https://github.com/openhab/openhab-core/issues/937#issuecomment-513843563).

The binding would be able to recover from issues such as:
* starting openHAB without the Z-Wave USB stick inserted and inserting it at a later time
* serial port permission issues that occur and are fixed while using the binding
* unplugging/reinserting the Z-Wave USB stick while using the binding
* other bundles providing serial comm dependencies going unavailable because of them being restarted/updated while using the binding

bodiroga:
Hi @cdjackson!

Let's see what you think about the implementation, fingers crossed :crossed_fingers:

Some parts of the code are taken from the great work done by @Hilbrand on #1210, so thanks Hilbrand for your job!

A couple of important notes:

- I have set the "notifyOnDataAvailable" flag to false because we are not using the serialEvent callback of the SerialPortEventListener interface to check for new data, as we are reading the serial port in a while loop. The "forever" sleep (```Thread.sleep(Long.MAX_VALUE);```) in serialEvent causes a 100% CPU load when the USB stick is removed, because the sleep thread continues to be alive even after interrupting and setting the receiveThread object to null. I don't know why, but I made some test setting the sleep to something like 300 seconds and logging a message after the sleep line, and after calling the onSerialPortError, when everything should be closed, the message was logged after 5 minutes. Adding the "```notifyOnDataAvailable(false)```" line has fixed the CPU load problem.
- Yes, I know that the "getSerialPortIdentifier" function is very awkward, because it's the same as the one defined in the serialPortManager interface, but using the already defined "getIdentifier" function I was not able to detect that the USB controller was unplugged. In my case the function does not return null if the stick was previously connected. It would be really helpful if you could make some tests overwriting the getSerialPortidentifier code with a simple ```return serialPortManager.getIdentifier(portId)```. Does the function return null when you remove the stick from the USB port?

Otherwise, the code is working flawlessly here and I have re-plugged the stick many times, trying to be a bad guy (when the network was initializing and the controller was sending a lot of messages), but everything restarts correctly. I also made the test of unplugging the controller, adding a new Z-Wave manually (with the sticks button) and plugging it back, and the new discovery result appeared in the inbox :+1:  

Thank you for having a look at it and looking forward to your comments. Best regards,

Aitor

Closes #1213


src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -80,49 +89,86 @@ public void initialize() {
...
+        } else {
+            return null;
+        }
+    }
+
+    private void watchSerialPort() {
         try {
-            SerialPortIdentifier portIdentifier = serialPortManager.getIdentifier(portId);
+            SerialPortIdentifier portIdentifier = getSerialPortIdentifier(portId);
             if (portIdentifier == null) {
```
cdjackson: I suggest that we make the port a required configuration in the thing XML file and then we can remove this check.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -80,49 +89,86 @@ public void initialize() {
...
+
+    /**
+     * Gets a serial port identifier for a given name.
+     * Workaround for getIdentifier in SerialPortManager class, because it doesn't detected correctly that the device is
+     * unplugged.
+     *
+     * @param the name
+     * @return a serial port identifier or null
+     */
+    private @Nullable SerialPortIdentifier getSerialPortIdentifier(final String name) {
```
cdjackson: I suggest that we make the port a required configuration in the thing XML file and then we can remove this method.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -80,49 +89,86 @@ public void initialize() {
 
         portId = (String) getConfig().get(CONFIGURATION_PORT);
 
-        if (portId == null || portId.length() == 0) {
+        if (portId.isEmpty()) {
```
cdjackson: I suggest that we make the port a required configuration in the thing XML file and then we can remove this check.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -131,6 +177,57 @@ public void initialize() {
...
+        logger.debug("Stopped ZWave serial handler");
+
+        super.dispose();
+    }
+
+    /**
+     * Stops resources when serial port is disconnected.
+     */
+    private void onSerialPortError(String errorMessage) {
+        logger.warn("Serial port error: '{}'", portId);
```
cdjackson: Please remove this logging event - it's not needed.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -131,6 +177,57 @@ public void initialize() {
...
+            serialPort = null;
+        }
+        logger.trace("Serial connection disposed");
+    }
+
+    /**
+     * Closes receive thread
+     */
+    private void disposeReceiveThread() {
+        logger.trace("Disposing receive thread");
```
cdjackson: Please make some of these messages debug level. I've not tried to run this, but some of these messages about closing down the serial connection should be raised so I can see what is happening. Currently they all seem to be trace.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -195,13 +286,16 @@ public void serialEvent(SerialPortEvent arg0) {
...
+                if (serialPort == null) {
+                    return;
+                }
+                synchronized (outputStream) {
+                    outputStream.write(response);
+                    outputStream.flush();
                     logger.trace("Response SENT {}", response);
                 }
             } catch (IOException e) {
-                logger.warn("Exception during send", e);
```
cdjackson: Please use warn level on this

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -240,7 +337,7 @@ public void run() {
                             continue;
                         }
                     } catch (IOException e) {
-                        logger.warn("Got I/O exception {} during receiving. exiting thread.", e.getLocalizedMessage());
+                        logger.debug("Got I/O exception {} during receiving. exiting thread.", e.getLocalizedMessage());
```
cdjackson: Please use warn

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -343,17 +440,19 @@ public void run() {
                     }
 
                 }
-            } catch (Exception e) {
-                logger.warn("Exception during ZWave thread. ", e);
+            } catch (RuntimeException e) {
+                logger.debug("Exception during ZWave thread. ", e);
```
cdjackson: Please use warn

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -365,13 +464,14 @@ public void sendPacket(SerialMessage serialMessage) {
...
-                serialPort.getOutputStream().write(buffer);
-                serialPort.getOutputStream().flush();
+            synchronized (outputStream) {
+                outputStream.write(buffer);
+                outputStream.flush();
                 logger.debug("Message SENT");
             }
         } catch (IOException e) {
-            logger.warn("Got I/O exception {} during sending. exiting thread.", e.getLocalizedMessage());
+            logger.debug("Got I/O exception {} during sending. exiting thread.", e.getLocalizedMessage());
```
cdjackson: Please use warn

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java
```diff
@@ -120,16 +120,22 @@ public ZWaveController(ZWaveIoHandler handler) {
     }
 
     public void shutdown() {
+        logger.trace("Shutting down ZWave controller");
```
cdjackson: Please use debug

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveTransactionManager.java
```diff
@@ -210,10 +213,30 @@ public ZWaveTransactionManager(ZWaveController controller) {
      * Shuts down the manager and frees resources
      */
     public void shutdown() {
+        logger.trace("Shutting down transaction manager");
```
cdjackson: please use debug

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveTransactionManager.java
```diff
@@ -432,7 +459,7 @@ public void run() {
                 try {
                     incomingMessage = recvQueue.take();
                 } catch (InterruptedException e) {
-                    logger.debug("Interrupted taking message from recvQueue", e);
+                    logger.debug("Interrupted taking message from recvQueue");
```
cdjackson: Please use trace

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveTransactionManager.java
```diff
@@ -995,6 +1028,10 @@ public void run() {
                 // Loop through all outstanding transactions to see if any have timed out
                 Iterator<ZWaveTransaction> iterator = outstandingTransactions.iterator();
                 while (iterator.hasNext()) {
+                    if (!running) {
+                        logger.trace("Stop looping through oustanding transactions"); // Change to trace
```
cdjackson: Please remove the comment

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveTransactionManager.java
```diff
@@ -1154,8 +1191,10 @@ public void transactionEvent(ZWaveTransaction transactionEvent) {
         try {
             ZWaveTransactionResponse response = futureResponse.get();
             return response;
-        } catch (InterruptedException | ExecutionException e) {
-            logger.debug("NODE {}: sendTransaction interrupted", e);
+        } catch (InterruptedException e) {
+            logger.debug("NODE {}: sendTransaction interrupted", transaction.getDestinationNode());
```
cdjackson: Please use trace

src/main/java/org/openhab/binding/zwave/internal/protocol/initialization/ZWaveNodeInitStageAdvancer.java
```diff
@@ -285,10 +321,11 @@ private boolean processTransaction(ZWaveMessagePayloadTransaction transaction, l
                 try {
                     Thread.sleep(backoff);
                 } catch (InterruptedException e) {
+                    logger.debug("NODE {}: processTransaction sleep interrupted", node.getNodeId());
```
cdjackson: Please use trace

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -190,6 +189,33 @@ protected void initializeNetwork() {
...
+        initializeHeal();
+
+    }
+
+    /**
+     * Common stopping point for all ZWave controllers.
+     * Called by bridges when their interfaces are unavailable.
+     */
+    protected void stopNetwork() {
+        logger.trace("Stopping ZWave network");
```
cdjackson: Please use debug

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -44,15 +51,22 @@
...
+    private static final long WATCHDOG_INIT_SECONDS = 5;
+    private static final long WATCHDOG_CHECK_SECONDS = 30;
+
     private final Logger logger = LoggerFactory.getLogger(ZWaveSerialHandler.class);
 
-    private SerialPortManager serialPortManager;
+    private final SerialPortManager serialPortManager;
 
-    private String portId;
+    private String portId = "";
```
cdjackson: This ="" can be removed if we add this as required to the XML

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
```diff
@@ -80,49 +89,86 @@ public void initialize() {
...
+            zWaveReceiveThread.start();
 
             // RXTX serial port library causes high CPU load
             // Start event listener, which will just sleep and slow down event loop
-            serialPort.addEventListener(receiveThread);
-            serialPort.notifyOnDataAvailable(true);
+            commPort.addEventListener(zWaveReceiveThread);
+            commPort.notifyOnDataAvailable(false);
 
             logger.info("Serial port is initialized");
```
cdjackson: This can probably be debug

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
bodiroga: Hi @cdjackson! I'm not sure if I understand your suggestion. If we want to use the ESH serial transport API, we need to get a SerialPortIdentifier object, so the call to the function is required. How should we otherwise open the serial port? :confused:

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
bodiroga: As the ZWaveSerialHandler class is now annotated with @NonNullByDefault, it is necessary to give the portId attribute an initial value, sorry.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
bodiroga: I have checked the controller xml file and the port parameter was already defined as "required", so yeah, I suppose that at that point the portId cannot be empty, although I would prefer if we could leave it as it is, just in case something goes wrong.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
bodiroga: Ummmm... wouldn't it be better if we log as debug the "```logger.info("Connecting to serial port '{}'", portId);```" message and leave this as it is? IMHO, the "is initialized" message is more meaningful than the "connecting to...". Anyway, as you wish :wink:

cdjackson: I don't think we need this message as a general message. We should really just log exceptions (ie things that don't go as expected) at info since the user is really only going to look at the log when there is a problem (IMHO).

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
cdjackson: Sorry - I misread what this was doing.

src/main/java/org/openhab/binding/zwave/handler/ZWaveSerialHandler.java
cdjackson: I don't see the need - it can never be null and adding the extra checks just makes the code less readable.

bodiroga:
Comments addressed, sorry for the late response!

It would be awesome if you could test the stick plug/unplug detection without the workaround, just returning ```return serialPortManager.getIdentifier(portId)``` in the ```getSerialPortidentifier``` function. I would like to know if the original function only behaves incorrectly for me or if it is a general problem. 

Thanks @cdjackson :wink:

cdjackson:
@bodiroga sorry it has taken so long - things have been very busy recently. I've tested this and once CI is complete will merge.

Thanks again for this well requested feature.

bodiroga:
You are more than welcome Chris, thanks for all your amazing work through this years :heart: 

PS: Next step: let's see if we can make the rfc2217 implementation work :wink:

PR #1241, opened 2019-10-09
Updated schema location from smarthome -> openHAB

Hilbrand:
Signed-off-by: Hilbrand Bouwkamp <hilbrand@h72.nl>


cdjackson:
Unfortunately there is no point in merging this since it will be overwritten in 2 days time when the database is updated. The change need to be made elsewhere.

wborn:
Did you already make this change elsewhere @cdjackson? If not, we should track it with an issue.

cdjackson:
I already made the change to the exporter.  Is this really needed to get merged quickly or not? I assume not, but if it is it will take me quite a lot of work to reexport the whole database again (I spent many days doing this a week or two back and don’t want to do this again for a few months).

wborn:
Thanks for the update. It's doesn't need to be merged quickly. It would be nice if it's done before openHAB 2.5.0 is released (probably somewhere in December).

cdjackson:
I wasn’t planning to do another database export before 2.5. The one I did last week was my 2.5 resync. Is this a “nice to have” or a “really required”. If it’s really needed I can try and do another in late November if I have time.

cdjackson:
Ok, thanks. I probably won’t do another full export then until after 2.5 to avoid potentially screwing up the database in the meantime. All new exports should have the new schema at least.

wborn:
Yes it's nice to have. I don't think it will fail the build when SAT is updated because it looks like it's configured not to:

https://github.com/openhab/org.openhab.binding.zwave/blob/bea7de6d78d5610a129150e2b9c90714967c7a02/pom.xml#L21

It would also be "nice to have" if you can update to indent the generated XML using tabs. That would solve many "OnlyTabIndentationCheck" warnings:

> There were whitespace characters used for indentation. Please use tab characters instead

Because fewer tabs chars are needed compared to spaces for indentation, it'll reduce the total size and improves XML parsing/generation times a bit.

cdjackson:
Updating to use tabs is really going to be a lot of work - this was all written 3 or 4 years ago (before someone decided to enforce tabs for XML and spaces for code) and there’s a lot of code that generates the XML files “by hand” that would take a lot of effort to update. If this is needed, then I can do it, but it’s quite a lot of work and I definitely won’t be able to do it for quite some time.

wborn:
I just wanted to mention it's "nice to have" so you could potentially solve both with the same export. As a bystander I never saw/used the export code so I make these comments without knowing how much work actually is involved in making these changes. :wink:

PR #1248, opened 2019-10-15
[infrastructure] Added spotless config, see openhab-core issue 935

seime:
See https://github.com/openhab/openhab-core/issues/935

Signed-off-by: Arne Seime <arne.seime@gmail.com>


cdjackson:
What file types does this format? If it formats XML then I'd suggest to remove this.

seime:
The current setup is to format java source code, pom.xml and xml files located in the src folder. 

The plugin supports ant style inclusion and exclusion of files, so which files (if any) that gets reformatted/style checked can be fine tuned.

cdjackson:
Please can you remove the XML then as this will result in 1000 files being reformatted backwards and forwards every time the formatter is run. Thanks.

seime:
IMHO, wouldn't it be better if they were reformatted to follow the code style defined by openHAB, and your editor was adjusted to follow this standard as well? 

If you are worried about performance, the automatic check could be adjusted to run only during CI builds (it does take a whopping 11s on my pc to reformat all plugin code)

cdjackson:
Yes, it probably would be better to reformat them, but the system that generates the files would need a lot of work and I don’t have time to do this at the moment.

seime:
Depends on https://github.com/openhab/openhab-core/pull/1101

seime:
Depends on https://github.com/openhab/org.openhab.binding.zwave/pull/1254

seime:
@cdjackson Ready for merge. 

The current setup does **not** change the code, but at a convenient time you should run `mvn spotless:apply` to reformat the current code (genererate thing xml files are currently excluded as requested). After that you should remove the `<spotless.check.skip>` property from the `pom.xml` `<properties>`section in order to enforce codestyle from that point and onwards.

cdjackson:
Thanks. So for my education (as I’ve not looked at this and am on holiday) this reformats the code during a maven build? or does the user have to explicitly run a maven goal to format code? And it should format exactly the same as the Eclipse formatter I guess?

seime:
When you remove the `<spotless.check.skip>` property (or change it to `true`)  it will *verify* that you follow the code standard - or fail the build. 

Reformatting is meant to be done by your IDE, but if it fails to do so you can also run `spotless:apply` to do it from Maven manually. In order to automatically *format* the code at build time you need to bind the `apply` goal to the lifecycle instead of the current `check` goal.

cdjackson:
Thanks - that sounds good.

PR #1271, opened 2019-12-21
Update ZWaveConfigProvider.java

mclassen2:
I add to 1 Minutes and 5 Minutes for Device Configuration of polling period make more options to me and other people too choice instead have how many seconds of polling period


cdjackson:
I don't really want to persuade people to use faster polling times. Polling is generally bad in ZWave and that's why I set 5 minutes as the minimum.

You can of course use whatever value you like, but I don't really see the need to provide this to others such that people think that it's acceptable.

PR #1285, opened 2020-01-04
Add notification_siren channel

cdjackson:


kennetn:
I did a test and I can confirm that the change works. I get a notification_siren3 and this is changing to ON when I activate the ringer and OFF when this tone stops playing.

Issue #1297, opened 2020-01-25
Binding transaction manager not waiting for report from node after UPDATE_NEIGHBORS


ilektron:
My reasoning that the transaction manager timeout is too short is this, 

```
2020-01-20 14:25:00.534 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 71: Configuration update received
2020-01-20 14:25:00.540 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 71: Configuration update set action_heal to true (Boolean)
2020-01-20 14:25:00.542 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 71: Starting heal on node!
2020-01-20 14:25:00.544 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 71: Starting network mesh heal.
2020-01-20 14:25:00.546 [DEBUG] [alization.ZWaveNodeInitStageAdvancer] - NODE 71: Starting initialisation from HEAL_START
2020-01-20 14:25:00.549 [DEBUG] [alization.ZWaveNodeInitStageAdvancer] - NODE 71: Node advancer - advancing to HEAL_START
2020-01-20 14:25:00.550 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 71: Got an event from Z-Wave network: ZWaveInitializationStateEvent
2020-01-20 14:25:00.551 [DEBUG] [alization.ZWaveNodeInitStageAdvancer] - NODE 71: Node advancer - advancing to UPDATE_NEIGHBORS
2020-01-20 14:25:00.553 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 71: Got an event from Z-Wave network: ZWaveInitializationStateEvent
2020-01-20 14:25:00.554 [DEBUG] [alization.ZWaveNodeInitStageAdvancer] - NODE 71: Node advancer: UPDATE_NEIGHBORS - updating neighbor list
2020-01-20 14:25:00.556 [DEBUG] [equestNodeNeighborUpdateMessageClass] - NODE 71: Request neighbor update
2020-01-20 14:25:00.557 [DEBUG] [nal.protocol.ZWaveTransactionManager] - NODE 71: sendTransaction org.openhab.binding.zwave.internal.protocol.ZWaveSerialPayload@1d2753e
[...]

PR #1299, opened 2020-01-25
Fix for issue #1297, failed update neighbors

ilektron:
Just increased the Z-Wave transaction manager's timeout for report requests to 10000ms instead of 5000ms.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveTransactionManager.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveTransactionManager.java
@@ -169,7 +169,7 @@ public class ZWaveTransactionManager {
     /**
      * Timeout in which we expect the request from the controller
      */
-    private final long timer2 = 5000;
+    private final long timer2 = 10000;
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks - I agree with your assessment of what you've seen, however I'm not keen on this implementation as it will have some nasty side effects. This will now double the timeout time where nodes don't respond which will cause problems with people who have non-responding nodes.

I would prefer to change the timer on a class by class basis - as it done elsewhere where we find that some commands respond slower. Please can you identify specific commands and update the ```doRequest``` method to increase the timer.

ilektron:
@cdjackson I completely agree with your assessment about it having an overly broad impact and possible side effects. I'll do some more poking to get it narrowed down and submit a new PR.

cdjackson:
Thanks.

You could have left this PR open and just make changes to it if you prefer - it keeps the conversation in one place so we remember in 12 months what we did :)

ilektron:
You have a point!

cdjackson:
Superseded by #1311

PR #1303, opened 2020-01-30
More concise binding description

ghys:
This is a proposal for a concise binding description which would be more in line with the other bindings.

Moreover, since it is not guaranteed that an UI will support HTML rendering of the description, better not include HTML tags in it.

Signed-off-by: Yannick Schaus <github@schaus.net>

### DIFF
--- a/src/main/resources/ESH-INF/binding/binding.xml
+++ b/src/main/resources/ESH-INF/binding/binding.xml
@@ -4,17 +4,7 @@
 	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
 
 	<name>Z-Wave Binding</name>
-	<description>
-	   <![CDATA[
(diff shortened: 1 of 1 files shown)


ghys:
image

cdjackson:
Isn't this a problem with the UI?

In the past (ok, many years ago) there was a discussion about using HTML so that we could have extended descriptions on bindings and parameter descriptions etc. The UI was only to render the first line and an ellipsis (or something) to allow the full description to be provided.

cdjackson:
I see in the description above that you are not actually processing HTML at all - I can see the ```<br>``` tokens in the description.

ghys:
The Z-Wave binding's description being detailed and multi-line while all the others' are only one or two (rather dull, admittedly) plain text sentences, it looked odd in that particular list and wondered why there was such an exception. There's an equivalent screen in Paper UI where it stands out as well.
In other places where it's displayed on its own, the HTML is processed.

Never mind, I'll find a way to make it work.

PR #1322, opened 2020-04-22
Add Property Type BINARY to INDICATOR_COMMAND_CLASS

Noobud:
I've do some work on the indicator command class to support Indicator_level property type binary.
Also I've bugfixed the indicator_command_class to support multiple endpoints.

The goal is to support the HeatIt Z-Scene Controller and other devices supporting this.

I've done some additions on the Z-Wave Device Database.
https://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/846

It is needed to correct the database export.

indicator_level
indicator_period
indicator_flash

this types need to include the configuration parameter(s)

indicator_period
indicator_flash

this types need to be exported as DecimalType.

indicator_level now supports an additional configuration parameter "property=BINARY". Default is MULTILEVEL, as this was the only option without this patch.

The complete story is in the community forum:
https://community.openhab.org/t/heatit-z-scene-controller-how-to-use-indicator-command-class/96887

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveIndicatorConverter.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveIndicatorConverter.java
@@ -68,7 +68,7 @@ public class ZWaveIndicatorConverter extends ZWaveCommandClassConverter {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks - Looks good to me.

Issue #1334, opened 2020-05-19
handle manufacturer proprietary report for fibaro fgr222


PatrikStaehli:
As several people mentioned in the forums, fibaro roller shutter state is not updated when operating in venetian blinds mode and being controlled manually (by physical switches).

* OH2 Version: 2.4.0
* Binding Version: >= 2.4.0 (also tested latest snapshot)
* Device name: FGR222
* Device version: 25.25
* Link to [database]: https://www.cd-jackson.com/index.php/zwave/zwave-device-database/zwave-device-list/devicesummary/413

Debug log:
```
[DEBUG] [WaveSerialHandler$ZWaveReceiveThread] - Receive Message = 01 0E 00 04 00 0B 08 91 01 0F 26 03 03 16 63 3A
[DEBUG] [nal.protocol.ZWaveTransactionManager] - processReceiveMessage input 0<>128 : Message: class=ApplicationCommandHandler[4], type=Request[0], dest=11, callback=0, payload=00 0B 08 91 01 0F 26 03 03 16 63
[DEBUG] [nal.protocol.ZWaveTransactionManager] - Received msg (0): Message: class=ApplicationCommandHandler[4], type=Request[0], dest=11, callback=0, payload=00 0B 08 91 01 0F 26 03 03 16 63
[DEBUG] [nal.protocol.ZWaveTransactionManager] - lastTransaction null
[DEBUG] [nal.protocol.ZWaveTransactionManager] - NODE 11: Application Command Request (ALIVE:DONE)
[DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 11: resetResendCount initComplete=true isDead=false
[DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 11: Incoming command class COMMAND_CLASS_MANUFACTURER_PROPRIETARY, endpoint 0
[DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 11: Command class COMMAND_CLASS_MANUFACTURER_PROPRIETARY not found, trying to add it.
[DEBUG] [tocol.commandclass.ZWaveCommandClass] - NODE 11: Unknown command class 0x91
[DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 11: Unsupported Z-Wave command class COMMAND_CLASS_MANUFACTURER_PROPRIETARY
[...]

PR #1335, opened 2020-05-19
fixed handling of manufacturer proprietary reports for fgr222

PatrikStaehli:
This is just a suggestion on how it could be fixed.
Please give feedback on how you wanted to go on with manufacturer proprietary stuff (ZWaveCommandClass and ZWaveManufacturerProprietaryCommandClass specifically).
Issue: https://github.com/openhab/org.openhab.binding.zwave/issues/1334

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCommandClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCommandClass.java
@@ -311,9 +311,7 @@ public abstract class ZWaveCommandClass {
             ZWaveEndpoint endpoint) {
         try {
             CommandClass commandClass = CommandClass.getCommandClass(classId);
-            if (commandClass != null && commandClass.equals(CommandClass.COMMAND_CLASS_MANUFACTURER_PROPRIETARY)) {
-                commandClass = CommandClass.getCommandClass(node.getManufacturer(), node.getDeviceType());
-            }
+
             if (commandClass == null) {
                 logger.debug("NODE {}: Unknown command class 0x{}", node.getNodeId(), Integer.toHexString(classId));
                 return null;
@@ -654,17 +652,6 @@ public abstract class ZWaveCommandClass {
             }
 
             return codeToCommandClassMapping.get(i);
-        }
-
-        /**
-         * Lookup function based on the manufacturer and device ID.
-         *
-         * @param manufacturer the manufacturer ID
-         * @param deviceId the device ID
-         * @return enumeration value of the command class or null if there is no command class.
-         */
-        public static CommandClass getCommandClass(int manufacturer, int deviceId) {
-            return getCommandClass(getKeyFromManufacturerAndDeviceId(manufacturer, deviceId));
         }
 
         /**
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveManufacturerProprietaryCommandClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveManufacturerProprietaryCommandClass.java
@@ -20,6 +20,8 @@ import org.openhab.binding.zwave.internal.protocol.ZWaveEndpoint;
 import org.openhab.binding.zwave.internal.protocol.ZWaveNode;
 import org.openhab.binding.zwave.internal.protocol.commandclass.impl.CommandClassManufacturerProprietaryFibaroFgrm222V1;
 import org.openhab.binding.zwave.internal.protocol.event.ZWaveValueEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 
@@ -32,6 +34,8 @@ import com.thoughtworks.xstream.annotations.XStreamAlias;
  */
 @XStreamAlias("COMMAND_CLASS_MANUFACTURER_PROPRIETARY")
 public class ZWaveManufacturerProprietaryCommandClass extends ZWaveCommandClass {
+    private static final Logger logger = LoggerFactory.getLogger(ZWaveManufacturerProprietaryCommandClass.class);
+
     private ManufacturerProprietaryClass classType;
 
     /**
@@ -61,6 +65,7 @@ public class ZWaveManufacturerProprietaryCommandClass extends ZWaveCommandClass 
 
     @ZWaveResponseHandler(id = 1, name = "PROPRIETARY_HANDLER")
     public void handleManufacturerProprietaryReport(ZWaveCommandClassPayload payload, int endpoint) {
+        logger.debug("handling manufacturer proprietary report");
         Map<String, String> values = CommandClassManufacturerProprietaryFibaroFgrm222V1
                 .handleFgrm222Report(payload.getPayloadBuffer());
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveManufacturerProprietaryCommandClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveManufacturerProprietaryCommandClass.java
@@ -18,6 +18,8 @@ import org.openhab.binding.zwave.internal.protocol.ZWaveCommandClassPayload;
 import org.openhab.binding.zwave.internal.protocol.ZWaveController;
 import org.openhab.binding.zwave.internal.protocol.ZWaveEndpoint;
 import org.openhab.binding.zwave.internal.protocol.ZWaveNode;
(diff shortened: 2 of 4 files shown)


cdjackson:
@PatrikStaehli thanks for looking at this.  I have a few comments that we need to think about before taking this further.

Firstly, it seems that the problem at the moment is there's a mismatch in type from the command class and the converter. You've elected to change the command class types, but I would prefer to change the converter. The command class is autogenerated code, so changing this manually causes problems.

Secondly, by removing the code that detects the device, you have effectively made this only usable by the Fibaro FGR222. Other devices also implement manufacturer proprietary commands, so they would now be incorrectly processed. We need to have a way to detect the device type, and that needs to be configurable externally (eg through the database) so when we add new types, we don't have to make Java code changes. Maybe this is broken, or not fully implemented in the current code (from my quick look that could be the case) - if that's the situation then we should fix it and not hard code the FGR222 implementation.

I hope that makes sense. If I've misinterpreted something, please just shout and we can discuss further.

PatrikStaehli:
@cdjackson thanks for your feedback - I was quite sure that this would not be the final solution - my goal was more to show that most of the code required is actually already there.

I do have quite a few questions - I'll try to only ask the most important ones for now:
1) which files are autogenerated (only the ones which actually say so, or all ZWaveCommandClasses)?
2) what are they autogenerated from? `CommandClassManufacturerProprietaryFibaroFgrm222V1` says it's autogenerated, but I could not find any hints on the source of it.
3) I don't get how manufacturer proprietary functionality should be implemented. If I understand the code correctly, for each device, there should be an instance of `ZWaveManufacturerProprietaryCommandClass` which receives the name of the actual device specific class via a setOptions call (option name is "className"). But that would not make sense because `ZWaveCommandClass.getInstance` tries to get the actual class name via manufacturer id / device type id, which only seems to query a static enum registry of basic class types but no hint on manufacturer proprietary classes? Of course, I can add `COMMAND_CLASS_MANUFACTURER_PROPRIETARY_FGR222(0x010f, 0x0302, "FIBARO_FGR_222",  ZWaveManufacturerProprietaryCommandClass.class)` to the CommandClass enum, but that seems to defeat the purpose of not wanting to write specific code for every proprietary functionality? Or is the idea to subclass `ZWaveManufacturerProprietaryCommandClass` for every device with proprietary code? But like that, the "className" option does not make sense any more...
4) Shouldn't there be an option called "className" in the database for the fgr222?
5) I can leave the obvously generated code unchanged (just changed the types in the converter), but I had to add another ZWaveResponseHandler to `ZWaveManufacturerProprietaryCommandClass`.

Once I know which code is auto generated from what source, I might get a better understanding on what should be done in the database (assuming this is the source of the autogenerated code) and what has to be coded by hand...

Kind regards
Patrik

cdjackson:
> which files are autogenerated (only the ones which actually say so, or all ZWaveCommandClasses)?

Only the ones that say so.

> what are they autogenerated from?

I have a code generator - I prefer not to change this just for the FGR222 as it will obviously break everything else.

> I don't get how manufacturer proprietary functionality should be implemented.

There needs to be some way to select the implementation from the database. I've not looked at this for a long time, but the idea is that there are multiple implementations for different devices, and this can be selected by a setting in the database.


> Shouldn't there be an option called "className" in the database for the fgr222?

Quite possibly - as above, I've not looked at this for a long time so it's quite possible. Maybe you can take a look through the code and see what is missing.

> I can leave the obvously generated code unchanged (just changed the types in the converter), but I had to add another ZWaveResponseHandler

Why is that? This should never be called - the correct logic is in the device specific class. We "just" need to select the correct class but you removed this code. Maybe the code didn't work correctly, in which case we should fix it - not remove it...

Cheers
Chris

PatrikStaehli:
> I have a code generator - I prefer not to change this just for the FGR222 as it will obviously break everything else.

No need to change the code generator - I've changed the converter and ZWaveValueEvent instead - but I still wonder what it generates the code from? The database? I don't see how eg `CommandClassManufacturerProprietaryFibaroFgrm222V1` could be generated from the database file `fgrm222_0_0.xml` .

> There needs to be some way to select the implementation from the database. 

Ok, fine. You probably mean the option I mentioned (`ZWaveManufacturerProprietaryCommandClass` expects an option with key "className").
It would still be necessary to extend `ZWaveManufacturerProprietaryCommandClass` for every device with manufacturer proprietary messages, so it can delegate the calls to the specific class:
```java
    @ZWaveResponseHandler(id = 1, name = "PROPRIETARY_HANDLER")
    public void handleManufacturerProprietaryReport(ZWaveCommandClassPayload payload, int endpoint) {
        Map<String, Object> values = null;
        switch (classType) {
            case FIBARO_FGRM222_V1:
                values = CommandClassManufacturerProprietaryFibaroFgrm222V1
                        .handleFgrm222Report(payload.getPayloadBuffer());
                break;
```
Is that what you expected, or did you have some kind of reflection based approach in mind? 

But there is also this code of `ZWaveCommandClass` which looks like it should select a manufacturer specific class derived from manufacturerId and deviceType (and I have yet to understand how this is supposed to work):
```java
    public static ZWaveCommandClass getInstance(int classId, ZWaveNode node, ZWaveController controller,
            ZWaveEndpoint endpoint) {
        try {
            CommandClass commandClass = CommandClass.getCommandClass(classId);
            if (commandClass != null && commandClass.equals(CommandClass.COMMAND_CLASS_MANUFACTURER_PROPRIETARY)) {
                commandClass = CommandClass.getCommandClass(node.getManufacturer(), node.getDeviceType());
            }
```
I don't see why there are two different mechanisms for selecting a device specific class?
If the information on which proprietary class `ZWaveManufacturerProprietaryCommandClass` should delegate to is set from the database via option "className", I don't see what the getInstance special handling should be used for. IMHO, `ZWaveManufacturerProprietaryCommandClass` could just be setup like any other CommandClass.

Concerning adding a ZWaveResponseHandler to `ZWaveManufacturerProprietaryCommandClass`:
There was only one handler registered with id = 0 (name = "DEFAULT_HANDLER", method name `handleManufacturerSpecificReport`, but the message contains a CommandClassCommand of 0x01, so I needed to add another handler with id = 1 (the code is visible above) which then delegates the processing to the Fgrm222 specific class. I don't see how that should work without this.
The message processing seems to work like this (please tell me if that's not how it should work):
1) ZWaveSerialHandler$ZWaveReceiveThread receives raw message
2) ZWaveTransactionManager gets the message, delegates to ZWaveNode
3) ZWaveNode.processCommand is called, resolves the ZWaveCommandClass - if not already available on the endpoint, the command class is created by calling ZWaveCommandClass.getInstance(...)  and registered with the endpoint for the next time it's needed
4) ZWaveCommandClass.handleApplicationCommandRequest gets the message, retreives the CommandClassCommand from the message and calls the appropriate `ZWaveResponseHandlerMethod` of the `ZWaveManufacturerProprietaryCommandClass` via reflection.
5) `ZWaveManufacturerProprietaryCommandClass` delegates message parsing / processing to the manufacturer proprietary class `CommandClassManufacturerProprietaryFibaroFgrm222V1` and notifies event listeners registered at the controller.

I'm sorry for asking that many questions, but I do have a hard time really understanding the key concepts behind the code. I can debug unit tests and enable debug logging on the installed OH, which helps, but it's also not very clear how much the setup for the tests has in common with 'the real thing'.

I can create a new PR which shows what I understood from your suggestions.

best regards
Patrik

PatrikStaehli:
Or maybe the `ZWaveManufacturerProprietaryCommandClass` should already be there and not have to be dynamically created?
I've just seen that the only place where 'setOptions' is called (which `ZWaveManufacturerProprietaryCommandClass` needs to know to which class it should delegate) is during initialization (`ZWaveNodeInitStageAdvancer.doStaticStages`).
So maybe I need to find out why that does not happen in the first place?
Do you have a hint on how I can force a new initialization? It seems that this is persistent, so it's not done again when the binding is restarted?

cdjackson:
> I don't see how eg CommandClassManufacturerProprietaryFibaroFgrm222V1 could be generated from the database file fgrm222_0_0.xml .

No - it has nothing to do with the database. I have a code generator that does this.

> Is that what you expected, or did you have some kind of reflection based approach in mind?

I think that is fine.

> IMHO, ZWaveManufacturerProprietaryCommandClass could just be setup like any other CommandClass.

I'm not sure what you mean. There must be a mechanism to select a non-standard class - it can't use the standard mechanism. If there are currently two mechanisms (Ive not had the chance to look recently) then you're right - we only need one. What we can't have is to just use the standard `ZWaveManufacturerProprietaryCommandClass` for all devices like we do for other command classes since it is device dependant.

> Do you have a hint on how I can force a new initialization? It seems that this is persistent, so it's not done again when the binding is restarted?

Do you mean how to update the thing after updating the thing definition in the XML? If so, then you need to delete the thing and add it again - this allows OH to re-read the the XML definition - otherwise it is persisted in storage. I know this isn't nice - I have proposed a mechanism to allow it to be updated, but it got rejected by ESH in the past.

PatrikStaehli:
@cdjackson 
I've now submitted another pull request #1342 with another solution.

PR #1341, opened 2020-05-31
RFC2217 improvements

bodiroga:
* Rework workaround so binding can use undiscovered RFC2217 ports
* Catch UnsupportedCommOperationException
* Improve RFC2217 reconnection logic

Also-by: Wouter Born <github@maindrain.net>


bodiroga:
@wborn, please, confirm that you agree with what I have done. I have squashed all commits in a single one, but left your original commit message and add my changes at the end. If there is anything wrong, tell it to me and I will make the required changes as soon as possible :wink:

Some notes:

- When a RFC2218 connection is in use, the ```enableReceiveTimeout``` function does not complete, and as a consequence, the ```read()``` function is synchronous. In this case, a -1 value means that the connection with the RFC2217 server is closed, so we stop the ZWaveReceiveThread thread and call the onSerialPortError function.
- I have also removed some "updateStatus" calls, as I have realized that they were not necessary.

Great team work! :+1:

cdjackson:
Thanks guys.

PR #1342, opened 2020-06-01
1334 support proprietary reports

PatrikStaehli:
Hello @cdjackson 

This is my second try at fixing / completing the code for handling manufacturer proprietary reports of fibaro fgr-222 roller shutters in venetian blind mode (parameter 3: reports type set to 1).
This approach does not delete any code and will support future manufacturer proprietary code.

The first try (PR) was this: #1335 
The issue is here: #1334 

The following needed to be changed for it to work:
1) a new property in the database for the things (it should work for fgrm-222 and fgr-222) to force CommandClass instantiation an being able to provide a custom option called `className`
2) an enum entry in `ZWaveCommandClass.CommandClass` for both devices mapping to `ZWaveManufacturerProprietaryCommandClass`
3) change the value type of `ZWaveValueEvent` from String to object to match the generated code and the existing converter code.
4) fix a bug in `ZWaveCommandClass.handleApplicationCommandRequest` which prevented handling of command requests by the default handler
5) implement the handling of manufacturer proprietary reports for fibaro fgrm-222 / fgr-222 devices in `ZWaveManufacturerProprietaryCommandClass` which delegates the work to a specific, generated class, depending on the database property `className` added above

I still think it's kind of unnecessary to having to configure which class should be used for delegation by the database property and also specifically adding a manufacturer/device type specific entry in the CommandClass enum. I can come up with reasons for one and the other, but not for having both:
a) using the enum would enable to subclass `ZWaveManufacturerProprietaryCommandClass` for different manufacturers or even devices, which can then delegate to the proper generated code class. For that, the database property would not be needed, as the code would already know where to delegate to.
b) using the database property allows to have all manufacturer proprietary code to be handled by `ZWaveManufacturerProprietaryCommandClass` (which delegates to different classes, depending on the database property). This does not need the CommandClass - it could be setup straight forward by mapping 0x91 directly to `ZWaveManufacturerProprietaryCommandClass`.
c) other than being super-flexible on how to implement future proprietary stuff, I can't see why we should do both a) and b)
But after all, one of the main reasons I struggled to get this to work was the lack of an example, aka existing code already having been implemented - I seem to be the first to use these mechanisms.
With this code, it should now be easier to understand what one needs to do to add more proprietary code... 

I'd greatly appreciate any feedback on this solution - I'm no professional java developer (I code in C# for almost 10 years now and C++ before that), so I could very well have coded something un-java-ish.

Best regards
Patrik

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveManufacturerProprietaryConverter.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveManufacturerProprietaryConverter.java
@@ -79,10 +79,10 @@ public class ZWaveManufacturerProprietaryConverter extends ZWaveCommandClassConv
         ZWaveValueEvent valueEvent = (ZWaveValueEvent) event;
         switch (channel.getUID().getId()) {
             case "blinds_lamella":
-                value = Integer.parseInt(valueEvent.getValue("LAMELLA_POSITION"));
+                value = (Integer) valueEvent.getValue("LAMELLA_POSITION");
                 break;
             case "blinds_shutter":
-                value = Integer.parseInt(valueEvent.getValue("SHUTTER_POSITION"));
+                value = (Integer) valueEvent.getValue("SHUTTER_POSITION");
                 break;
         }
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCommandClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/commandclass/ZWaveCommandClass.java
@@ -267,16 +267,16 @@ public abstract class ZWaveCommandClass {
         }
 
         logger.debug("NODE {}: Received {} V{} {}", getNode().getNodeId(), getCommandClass(), getVersion(),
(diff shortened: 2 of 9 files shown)


PatrikStaehli:
Hello @cdjackson 
Were you able to have a look at it?
I'm aware that there don't seem to be a lot of people waiting for this, but it would still be nice to get this into the main repo sometime.
Please let me know if you want me to change something.
Best regards
Patrik

mheiss:
Is there any update on this pull-request? Are there open points that need to be addressed? Is help required?
I have about 10 of those devices and i would be happy if operating manually is correctly reflected in the UI.

PatrikStaehli:
I'm not aware of any open points. I've kept this change as close to the existing code as possible and it would be great to see it merged. Feel free to review the code if you like - every feedback is welcome.
Maybe @cdjackson could find some time to have another look at it?

PatrikStaehli:
I'll close this PR, as I switched to using zwave-js-ui (which supports proprietary fibaro reports) to control my devices through mqtt and I'm transitioning to home assistant anyway.
The code is there, if anyone should care to use (or even merge) it.
Regards
Patrik

PR #1349, opened 2020-06-30
Add textual configuration sample

theiding:
Samples are very powerful to quickly get users to understand the configuration. Removed the table that referred to the old device type. 

Signed-off-by: Mark Theiding <mark.theiding@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,16 +105,17 @@ There are a large number of things supported by the Z-Wave binding, so configura
 
 #### Textual Thing Configuration
 
-Things configured manually through text files require the following minimum configuration to be set. -:
-
-| Configuration      | Description                                                                                                   |
-|--------------------|---------------------------------------------------------------------------------------------------------------|
-| zwave_nodeid       | Sets the node id of the node within the network.                                                              |
-| zwave_manufacturer | Sets the manufacturer ID for this device (as decimal). This is used to get the thing type from the database.  |
-| zwave_deviceid     | Specifies the device ID for this device (as decimal). This is used to get the thing type from the database.   |
-| zwave_devicetype   | Specifies the device type for this device (as decimal). This is used to get the thing type from the database. |
-| zwave_version      | Specifies the application version for this device. This is used to get the thing type from the database.      |
-
+Things configured manually through text files must define a Bridge for the controller and all devices under that. 
+Follow this sample format:
+```
+Bridge zwave:serial_zstick:controller "ZWave Controller" [ port="/dev/ttyACM0", controller_softreset="false", controller_master="true", heal_enable="true", security_networkkey="XXX" ]
+{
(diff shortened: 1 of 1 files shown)


theiding:
This update was motivated by https://community.openhab.org/t/zwave-things-syntax/101561. 
Please note that I am linking to @cdjackson's very helpful community post - is that acceptable?

cdjackson:
Thanks for this Mark - I agree that examples are good :)

I would suggest not to remove the table of configuration - while the example is nice, I still think that an explanation of what the config parameters do is useful rather than just having the example.

I would not reference the forum - I'd suggest to make the document stand-alone.

theiding:
That makes sense. You support two ways of defining a channel - as specified in the example via device type or via the generic "device" device plus db entries, right? The table is referring to the latter if I understand correctly. I thought the latter approach was deprecated (and that's why I removed the table). If it is still relevant then should we have the table plus an example for that alternative format? 
Not referencing the forum is definitely much cleaner - but I didn't want to steal content from the forum. How about I lift the essential information - with your permission - from your forum post into the docs? (I remember from when I used your post a couple years back that I found your description on identifying the device type particularly helpful).

theiding:
@cdjackson , I reviewed the binding implementation and now I think I understand the configuration better. Here is what I saw:
- There is only one way to define a thing through a text file (using the format provided in the example). 
- Using the format "zwave:device..." in a text file simply defines the "Unknown Device". Using the format "zwave:device..." plus config parameters for manufacturer etc. doesn't work (since the framework has no mechanism to include config parameters and binding input in determining available channels). And there is no need to support this format for textual definition in the first place. 
- Discovered things are assigned a UID with the format "zwave:device...". It would be preferable to use the format using a concrete type (such as zwave:aeon_zw100_01_008...), but that is not possible since we want to put a discovered thing right away into the Inbox and not wait until the concrete type is discovered later, so we need to lock in the UID right from the beginning without the concrete type. 

I think the concrete type vs "device" often causes confusion  - I certainly wasn't clear. Assuming I got the above right I then propose the following doc update:
- explain the textual format with an example
- mention that "zwave:device..." is not supported for textual configuration
- lift your relevant details from your community post into the textual configuration section
- keep the table of manufacturer properties etc. but since they are not related to textual configuration move it from the textural configuration section to the discovery section

Let me know if this sounds right and I'll submit a revised PR.

cdjackson:
Thanks @theiding - I think this looks ok. My only comment is that the table of "manufacturer properties" is not a table of manufacturer properties - it's a bit of a mixture of properties and configuration. You need to leave the nodeid parameter there since this is a) not a property, and b) it's required to be configured.

theiding:
@cdjackson, finally got to adjust the PR. Hope this all makes sense now.

cdjackson:
Thanks.

PR #1351, opened 2020-07-05
Add shutdown of executor to avoid leaks if binding is restarted

seime:
Also added thread names to ease debugging


cdjackson:
Thanks. LGTM.

PR #1391, opened 2020-08-24
Add ZWave controller properties

### LINKED ISSUE #1390: Add some properties to the Thing of the controller
Hi @cdjackson!

As the binding reads several properties of the controller during its initialization, I think that we should add them to its Thing. This are the properties that could be added:

- Manufacturer ID
- Model ID
- Node ID
- Home ID

The first two could be used in the future to dynamically create a channel that allows the user to switch off the integrated LED light of the ZStick Gen 5. At the same time, the Home ID could be marked as the representation property of Thing (I still haven't given up on making the binding compatible with the USB discovery process :rofl:). Yes, I know that the Home ID is not unique and that it can be changed, but: 1) it is not so common to reset the controller and 2) if you have done it, you will have to delete all the Things you would have created, because those devices have to be included again.

I will provide a PR right now, so that we can discuss the actual implementation :wink: 

Best regards,

Aitor

bodiroga:
This PR adds the modelId, the manufactureId, the homeId and the nodeId as properties to the Thing of the serial controller. The homeId is set as the representation property of the Thing. Also, it marks all the Thing's channels as advanced, because they are not very used (or not used at all by non-expert users).

Closes #1390 

### DIFF
--- a/src/main/resources/ESH-INF/thing/controller_serial.xml
+++ b/src/main/resources/ESH-INF/thing/controller_serial.xml
@@ -17,6 +17,13 @@
             <channel id="serial_cse" typeId="serial_cse" />
         </channels>
 
+        <properties>
+            <property name="modelId"></property>
+            <property name="manufacturerId"></property>
+            <property name="nodeId"></property>
+        </properties>
+        <representation-property>homeId</representation-property>
+
         <config-description>
             <parameter-group name="port">
                 <label>Port Configuration</label>
@@ -198,7 +205,7 @@ In this mode, only <i>Entry Control Devices</i> such as locks will be securely i
         </config-description>
     </bridge-type>
 
-    <channel-type id="serial_sof">
+    <channel-type id="serial_sof" advanced="true">
         <item-type>Number</item-type>
         <label>Start Frames</label>
         <description>Counter tracking the number of SOF bytes received</description>
@@ -206,7 +213,7 @@ In this mode, only <i>Entry Control Devices</i> such as locks will be securely i
         <state pattern="%d" readOnly="true"></state>
     </channel-type>
 
-    <channel-type id="serial_ack">
+    <channel-type id="serial_ack" advanced="true">
         <item-type>Number</item-type>
         <label>Frames Acknowledged</label>
         <description>Counter tracking the number of frames acknowledged by the controller</description>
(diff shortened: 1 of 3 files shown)


src/main/resources/ESH-INF/thing/controller_serial.xml
```diff
@@ -17,6 +17,13 @@
             <channel id="serial_cse" typeId="serial_cse" />
         </channels>
 
+        <properties>
+            <property name="modelId"></property>
```
cdjackson, line 21: Is there a reason to add the properties here? I guess it doesn't hurt, but I'm wondering why do it?

bodiroga: Hi @cdjackson! Is your question only related to the "modelId" property or to all of them? I have created the PR because I would like to make use of the Home ID property in https://github.com/openhab/org.openhab.binding.zwave/pull/1388: reading the value through Python in the rfc2217 gateway (https://github.com/bodiroga/rfc2217-gateway) and announcing it using mDNS. Having the Home ID set as the representation property, the thing could be updated automatically if the gateway changes its IP address. I would update #1388 if this PR gets merged, sure. Anyway, I can delete the "modelId", "manufacturerId" and "nodeId" if you consider them superfluous, but the "homeId" property is essential to :wink:

cdjackson: My question is really - why did you need to put them in the XML file? I guess it doesn't hurt, but I don't list other properties, so it makes it a bit non-standard. I guess we could alternatively add other properties as well to make it standard - I'd just prefer one or the other and I'm not sure I see the point of adding it?

bodiroga: Hi Chris! I added them following the developers documentation: https://www.openhab.org/docs/developer/bindings/thing-xml.html#properties. But reading the following sentence again: ```Note that the representation property is normally not listed in the properties part of the Thing type XML, as this part contains static properties, that are the same for all instances of this Thing type```, I think that you are totally right. As the properties are dynamic and not the same for all serial Things (we have different serial controllers), I will remove them from the XML file, leaving the ```representation-property``` line alone. Sorry for my error :wink:

cdjackson:
Having not used the representation property before, I'm a bit unclear of its use. Does it need to be provided as part of the discovery? At the moment, this won't be the case and I'm a little unsure why it's really needed?

bodiroga:
Hi @cdjackson!

> At the moment, this won't be the case and I'm a little unsure why it's really needed?

I need the ```representation-property``` for https://github.com/openhab/org.openhab.binding.zwave/pull/1388, but first I need to update that PR since I'm really close to implement the reading of the Home ID in my https://github.com/bodiroga/rfc2217-gateway project. The gateway will read the Home ID of the stick and announce the information through mDNS, so that the discovery service can use it to set the thingId and the homeId property.

Perhaps I can remove it from here and add it to the other PR? Seems to be a good idea :+1:

bodiroga:
I have updated the PR to remove the representation property and I have also created some constants for the new properties :wink:

PR #1447, opened 2020-11-23
Extend Door Lock Command Class

DanielKnoop:
This PR adds two features to the door lock command class:
- Adds the capability to the door lock command class to accept UNKNOWN lock state, mapped to UndefType.UNDEF
- Adds the capability to handle BOLT and LATCH to the door condition event. Needs to be defined by a type in the device database.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveDoorLockConverter.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveDoorLockConverter.java
@@ -100,7 +100,7 @@ public class ZWaveDoorLockConverter extends ZWaveCommandClassConverter {
     }
 
     private State handleEventCondition(ZWaveThingChannel channel, ZWaveCommandClassValueEvent event) {
-        if (!channel.getUID().getId().equals("sensor_door")) {
+        if (!channel.getUID().getId().startsWith("sensor_door")) {
             return null;
         }
         Integer eventValue = (Integer) event.getValue();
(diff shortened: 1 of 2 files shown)


src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveDoorLockConverter.java
```diff
@@ -92,13 +100,24 @@ private State handleEventLockState(ZWaveThingChannel channel, ZWaveCommandClassV
     }
 
     private State handleEventCondition(ZWaveThingChannel channel, ZWaveCommandClassValueEvent event) {
-        if (!channel.getUID().getId().equals("sensor_door")) {
+        if (!channel.getUID().getId().startsWith("sensor_door")) {
```
cdjackson, line 103: Why have you changed this? There are no other such channels with this name. Probably this should be using the channel type anyway.

cdjackson (commented):
Thanks. I think in general this looks fine - just one small comment that probably needs fixing.

src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveDoorLockConverter.java
DanielKnoop: Yes I think that should be rolled back. It was a fix idea I do this morning, the other code was tested with my lock the whole weekend.

DanielKnoop: type should be better.
Is this channel.getChannelTypeUID().getId()?
Then I will modify it...

cdjackson: Yes, I think this should be best, but please try it to be sure :)

DanielKnoop: Hi Chris, 
yes, I've tested. It is productive since Monday. 
Daniel

cdjackson:
Ah - ok, I hadn't noticed that you'd updated the PR. Thanks.

PR #1461, opened 2020-12-14
Database update

cdjackson:


kaikreuzer:
@cdjackson Did you merge those updates to 2.5.x accidentially or was that on purpose?
I just noticed that there were no updates on the master branch for the last weeks, so it seems that the 3.0.0.RC1 build contains an outdated version. Or am I missing something?

cdjackson:
It was not accidental. The import to this branch is automatic and was required by a user using 2.5 - as was discussed with the user on the forum.

I’ll do an OH3 update shortly - this is also an automated script.

kaikreuzer:
Ok, perfect - thank you!

Issue #1531, opened 2021-02-10
[Database]


bwosborne2:
Database updates seem to be working for 2,5,x but not main or 3.x branches.

The Zooz ZEN34 xml is out of date here compared to the database.

cdjackson:
Database updates are only done for the master and 2.5 branches - not the 3.x branch. I’ve not checked specifically, but it strange if they are working for one branch then I don’t really know why it would have stopped on another as it’s all running from fundamentally the same script.  I will take a look at this toward the end of next week once the shippers have been.

bwosborne2:
It appears they got politically correct and renamed `master` to `main`

cdjackson:
What do you mean? Do you mean that the repository has changed the name of the master branch? I’ve not seen any issue or notes on this in the repository so that’s rather annoying if that’s happened!!! Crazy.

I will look at this in a week or so - at the moment I’m up to my neck in shipping details with exactly 1 week before shippers arrive :(

bwosborne2:
![image](https://user-images.githubusercontent.com/51490785/107513813-c75ce900-6b76-11eb-9fc5-71837e970b70.png)

bwosborne2:
Would it make sense for me to change the scripts and try?

cdjackson:
I don’t know who changed that, but it seems crazy to make such a fundamental change without any consultation with repository maintainers at least!

bwosborne2:
I just searched and, although new repos get that naming, I saw no indication of an automatic change. Just an indication of the crazy world today.

Roagert:
Yeah im missing a device that is listed in the opensmarthouse database. I could really use a database update. :) 
Sounds strange that they would make such a big change

bwosborne2:
I hope to take a look at it this evening.  I have a couple of ideas top at least get it more current.

Roagert:
Im actually still missing this device to be update strangely enough.
https://opensmarthouse.org/zwavedatabase/1000

bwosborne2:
> Im actually still missing this device to be update strangely enough.

Not too strange. That device was only approved on 2/9, after the export the PR above is based on.
There are still 3.x update issues and real life is adjusting priorities. Are you running version 3.x or 2.5.x ? 
This device MAY be in the latest 2.5.x snapshot.

Roagert:
> > Im actually still missing this device to be update strangely enough.
> 
[...]

PR #1532, opened 2021-02-10
manual database update

bwosborne2:
Manual database update.
Process scripts broke due to renaming from "master" to "main".

### DIFF
--- a/src/main/resources/OH-INF/thing/abus/shlm10000_0_0.xml
+++ b/src/main/resources/OH-INF/thing/abus/shlm10000_0_0.xml
@@ -24,7 +24,7 @@ ABUS Security Center Bulb<br /> <h1>Overview</h1><p>The ABUS Z-Wave LED lamp can
       <channel id="color_color" typeId="color_color">
         <label>Color Control</label>
         <properties>
-          <property name="binding:*:HSBType">COMMAND_CLASS_SWITCH_COLOR</property>
+          <property name="binding:*:HSBType">COMMAND_CLASS_SWITCH_COLOR;colorMode=RGB</property>
           <property name="binding:*:PercentType">COMMAND_CLASS_SWITCH_MULTILEVEL</property>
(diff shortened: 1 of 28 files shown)


bwosborne2:
This is a temporary workaround for issue #1531 
It is NOT a complete fix.

cdjackson:
Thanks Bruce.

bwosborne2:
I updated the script branch & tried from the web menu. It did not work and, I assume I would need to download the scripts & run locally.

I figured this could buy some time.

cdjackson:
As I’ve mentioned in the past, you can’t trigger the update from the web ui - that doesn’t work as there needs to be a decoupling between the UI and the script. It can only be done through ssh.

cdjackson:
Bruce - I tried to run your script, but get file not found error now. I'll try and take a look later tonight but did you rename it or something?

bwosborne2:
All I did in the OH3 script was change `master` to `main`

cdjackson:
Ok, this file seems to be gone now so I'll need to regenerate it as I don't think I have it elsewhere - I don't suppose you had a copy?

bwosborne2:
I just used the online editor. I do not know how it was deleted. It should be the same as the OH2 one except for the branches, correct?

cdjackson:
Ok, I'm not sure how it got deleted :(

I have a feeling there were another couple of small changes to make it work on OH3, but I can't remember now as it implemented 6 months back. Probably the folders also have to change for starters - I forget what else.

I won't be able to look at this for a week or so.

bwosborne2:
I was just thinking it may be in a backup.
I just tried the 3.1.0 SNAPSHOT with a 3.0.0- snapshot serial ( I had it available) on 3.0.1 Docker and it complains about the zsmoke xml :(
Since it is Docker, going to try 3.1.0 SNAPSHOT OH

PR #1551, opened 2021-03-04
Convert commands to channels datatype if possible

### LINKED ISSUE #1546: Eurotronic Spirit Thermostatic Valve sensor_report doesn't accept QuantityType
As already mentioned in a ticket on opensmarthouse.org the `sensor_report` channel for the Eurotronics Spirit Thermostatic Valve doesn't accept `QuantityType`s within openHAB 3.

For me it looks like the channel is not defined as `Number:Temperature`, but just `Number`. I would like to use the follow-profile to send a `Number:Temperature` to this devices sensor_report channel, but it doesn't work.

But @cdjackson already answered within the ticket, that my assumption is not correct and some changes within the binding sources are required. Therefore I am opening this issue here.

Log:

```
2021-01-25 21:19:03.305 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 62: Command for unknown channel zwave:device:controller:node62:sensor_report with QuantityType
2021-01-25 21:19:03.304 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 62: Command received zwave:device:controller:node62:sensor_report --> 17.2222222222222222222222222222222 °C [QuantityType]
```

I am not used to the sources of this binding, if someone gives me a hint how to fix this issue I can try it and do the changes.
Maybe just something like converting a QuantityType to DecimalType if target channel is defined as `Number` is required?

* OH2 Version: 3.0.0
* Binding Version 3.0.0
* Device name: Eurotronic Spirit Thermostatic Valve
* Device version: ?
* Link to [database](http://opensmarthouse.org/zwavedatabase): https://opensmarthouse.org/zwavedatabase/710
* If the device is not in the database, provide a link to the manual: Device already in DB
* If there is an XML being generted for the device (in the /userdata/zwave folder), please provide it: Doesn't seem to be required for this case, XML was provided within the ticket

chris922:
Fixes #1546 

I really don't know if this is the correct approach to solve this issue.. I am running this right now on my local instance it it seems to work.

The problem is that for example a `QuantityType` can't be send if there is no channel that expects a `QuantityType` even if e.g. there is a channel with a `DecimalType` to which a `QuantityType` can easily be converted to.

Thus I try to find a channel to which the source type (`QuantityType` in the example) can be converted to.

My approach:
* Stick to as-is behavior if a channel with the required datatype can be found and use this one
* If no channel with matching datatype can be found iterate over all channels with the same `channelUID` and try to convert the command to the one matching with the expected datatype, the first one found that can be converted will be used

Another idea I had:
Add a channel to this device that expects a `DecimalType` -> would most likely just fix the issue I am facing for this particular device and not similar issues for other devices

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -31,9 +31,11 @@ import java.util.TimeZone;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
+import org.jetbrains.annotations.Nullable;
 import org.openhab.core.config.core.Configuration;
 import org.openhab.core.config.core.status.ConfigStatusMessage;
 import org.openhab.core.config.core.validation.ConfigValidationException;
+import org.openhab.core.library.types.*;
 import org.openhab.core.thing.Bridge;
 import org.openhab.core.thing.Channel;
 import org.openhab.core.thing.ChannelUID;
@@ -1151,7 +1153,7 @@ public class ZWaveThingHandler extends ConfigStatusThingHandler implements ZWave
 
         DataType dataType;
         try {
-            dataType = DataType.valueOf(command.getClass().getSimpleName());
+            dataType = DataType.fromTypeClass(command.getClass());
         } catch (IllegalArgumentException e) {
             logger.warn("NODE {}: Command received with no implementation ({}).", nodeId,
                     command.getClass().getSimpleName());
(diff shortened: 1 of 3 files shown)


src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
```diff
@@ -1202,6 +1222,44 @@ public void handleCommand(ChannelUID channelUID, Command command) {
         }
     }
 
+    @Nullable
+    Command convertCommandToDataType(ChannelUID channelUID, DataType channelDataType, Command command,
+                                     DataType dataType) {
+
+        if (!(command instanceof State)) {
+            logger.debug("NODE {}: Command of data-type {} received for channel {} that must be converted to " +
```
cdjackson, line 1230: Could you look at shortening this debug message? It's a debug level, so doesn't really need to be so long. 

```
            logger.debug("NODE {}: Command of type {} received for channel {} to be converted to type {}, is not a State",
```

cdjackson (commented):
I generally think this is fine - thanks. There are a few debug messages that could be a bit shorter and more succinct - I commented on one, but maybe just look over them to see if they can be made more "punchy" :)

src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
chris922: Thanks for the feedback! In general I always spam debug logs, but you are right.. for this binding many people will have debug activated and it already produces a lot log messages.

I shortened the debug messages, is it now fine for you or did I removed too much information? :smile:

cdjackson: Thanks - looks good to me.

cdjackson:
Thanks @chris922 - looks good.

PR #1558, opened 2021-03-14
Remove org.jetbrains.annotations.Nullable

wborn:
Only Eclipse JDT annotations should be used, see:

https://www.openhab.org/docs/developer/guidelines.html#null-annotations


wborn:
This will fix the build @openhab/z-wave-maintainers!

PR #1573, opened 2021-04-05
Added Blinds control channel (tested)

pawelrojek:
I have Fakro ARZ Roller Shutters and I couldn't control them without blinds_control channel. Switch Dimmer also works, but it's very inconsistent (doesn't stop where it's supposed to).


cdjackson:
Please update the database at opensmarthouse.com rather than submitting a PR here. 

Thanks

Sent from my iPhone

PR #1577, opened 2021-04-06
Fix NullPointerException.

adrianboimvaser:


cdjackson:
Thanks. Please can you rebase this on main please. I don't believe there are any further builds of OH2 being made.

adrianboimvaser:
Not even maintenance, bug-fix builds?

cdjackson:
I’ve been keeping the database updated, but I noticed that CI is no longer working for all OH2.5 builds - not just ZWave. Until that is resolved, ZWave can’t be built even.

cdjackson:
So, I would also be happy to merge this - so long as you also create hte PR for OH3. As above - OH2.5 isn't being built now (you can also see that on this PR with CI) but I'm happy to merge in case one day 2.5 artifacts are built again.

PR #1590, opened 2021-04-22
Database update

cdjackson:


bwosborne2:
I fear you are going to have to build this manually.

cdjackson:
Sorry Bruce - I don't understand. This is just a PR - the building is done elsewhere. Any builds will have to wait till the next release - as is normal. I don't know what the release schedule is for this branch since I only just found out about it. I will try and do regular updates of the branch so that if there is a release, it is reasonably up to do.

bwosborne2:
I view things from an end user and end user support perspective. From that perspective, the update is not useful unless it has been built and available for installation. I do not see any 3.0.x build chains in Jenkins and do not know 3.0.2 was built and released. 

Ultimately that should be somebody else's concern.

cdjackson:
> I view things from an end user and end user support perspective. 

That's fine, but I can't do anything about that here. The distribution is managed via a different system, and different repositories.

Of course as with any update, it is not useful until it is built and released - that is always the case, but PRs are here to resolve issues, and the build and release will be done later to release the executable. That is exactly the same here - you can build manually as you suggest, or wait for the next release - as per normal.

PR #1604, opened 2021-05-09
added channel types meter_kvar & meter_kvarh

BorisFierlings:
Added channel types meter_kvar & meter_kvarh. Will be needed for updated Qubino smart meter database defintion.

Signed-off-by: Boris Fierlings <Boris.Fierlings@web.de>

### DIFF
--- a/src/main/resources/OH-INF/thing/channels.xml
+++ b/src/main/resources/OH-INF/thing/channels.xml
@@ -468,11 +468,11 @@
         </state>
     </channel-type>
 
-    <!-- Energy - Current Consumption -->
+    <!-- Energy - Current -->
     <channel-type id="meter_current">
         <item-type>Number</item-type>
-        <label>Electric Current Consumption</label>
-        <description>Indicates the instantaneous current consumption</description>
+        <label>Electric Current</label>
+        <description>Indicates the instantaneous current</description>
         <category>Energy</category>
         <state pattern="%.1f" readOnly="true">
         </state>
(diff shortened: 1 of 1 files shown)


src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -522,7 +532,7 @@
     <channel-type id="meter_kwh">
         <item-type>Number</item-type>
         <label>Energy Consumption</label>
-        <description>Indicates the energy consumption (kWh)</description>
+        <description>Indicates the instantaneous energy consumption (kWh)</description>
```
cdjackson, line 535: For me, this is wrong - you don't have instantaneous energy consumption measured in watt-hours. Energy consumption is measured in Wh -  but instantaneous power consumption is measures in watts.

I'm not sure how this links to the ZWave data definition and how that is applied to this channel, but fundamentally it's not correct to add instantaneous here and call it kWh.

src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -616,6 +626,16 @@
         </state>
     </channel-type>
 
+    <!-- Energy - Power Consumption -->
+    <channel-type id="meter_kvar">
+        <item-type>Number</item-type>
+        <label>Electric Power Consumption</label>
```
cdjackson, line 632: Suggest to use consistent naming here - you added "Energy Consumption" above but "Electric Power Consumption" here. Either use "Electric" in both or neither.

src/main/resources/OH-INF/thing/channels.xml
BorisFierlings: You are right. Togehter with energy consumption there is rarely used "instantaneous" but often "active" in opposite to the reactive and aparent energy consumption. So we can leave it as energy consumption or change to active energy consumption.

src/main/resources/OH-INF/thing/channels.xml
BorisFierlings: You are right again. As the "Electric" is not only in "Electric Power Consumption", but also in "Electric Current", it should be "Electric Energy Consumption". Sorry for that.

BorisFierlings:
Sorry for beeing off that long time. I've done the changes a suggested.

PR #1619, opened 2021-06-15
Corrected missing switch_binary3

BAMLABS-AU:


BAMLABS-AU:
## Expected Behavior
Upon successful installation of this thing, 3x switch_binary channels
should be available as this is the 3-Way switch version.

## Current Behavior
Only 2x Switch channels showing up to configure as items. 
binary_switch 1 and 2 show up. However binary_switch 3 is not available.

Updated the XML of the template that i could locate and sent for review.

cdjackson:
Please update any errors directly in the database. There should be a link from the doc for each device.

PR #1621, opened 2021-06-22
create callback ID so Add Node stop command doesn't time out

apella12:
This command is timing out in 5 seconds and reducing the configuration time for the node before it is put to sleep.  Shortening the time gets it out of the way. (This is my first try at creating a pull request, so be easy on me).  This is the background
https://community.openhab.org/t/wake-up-no-more-information-command-to-battery-node-not-getting-through/120139

I have a debug with this change below that configures the device in 5 seconds versus the other logs in the above link.[zwave.log](https://github.com/openhab/org.openhab.binding.zwave/files/6690179/zwave.log)

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -81,7 +81,7 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
 
         // Create the request
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
-                .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
+                .withPayload(ADD_NODE_STOP).withTimeout(300).withRequiresData(false).build();
 
         if (complete) {
             payload.setCallbackId(0);
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -81,10 +81,10 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
 
         // Create the request
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
-                .withPayload(ADD_NODE_STOP).withTimeout(300).withRequiresData(false).build();
+                .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
 
         if (complete) {
-            payload.setCallbackId(0);
+//            payload.setCallbackId(0);
         }
         return payload;
     }


apella12:
git commit --amend --no-edit --signoff

apella12:
Signed-off-by: roberteckhoff roberteckhoff@yahoo.com

cdjackson:
This doesn't seem like a good approach- you probably should try to find the problem and fix the bug properly rather than just reducing the timeout as the error is still going to be there.

apella12:
That's fine, since I wanted to try this PR thing and all my checks have failed anyway.  

I do believe the problem is that the command is set-up to need a Request (other parameters of 4A do need to contact the node) and from the Silabs documents it appears all that is needed is the the controller Ack.  I'm thinking that the node does not need to know that the Add Node is finished.  I can see from the Zniffer that the controller is not sending anything to the node, so nothing is going to come back and it is going to time out anyway, just trying to get it out of the way.

Thought I would try since battery devices need the extra 4 seconds.  Thanks for considering

cdjackson:
Do not delete the command - it is required. Also, the controller should definately send a response, so if it isn't then there is likely something wrong. As a minimum there should be an ack.

I don't currently have a ZWave system running as I'm just moving in to my house and getting things set up so can't easily test this, but I don't recall such an issue in the past.

apella12:
So should not try anything new like this PR so late at night.. 

I am editing through VSC as it shows some errors of my ideas. I remembered last night that I did try to suppress the command in question earlier and there were dependencies in other classes, so abandoned that idea.

To clarify the Controller does Ack, so the command moves from Wait Response to Wait Request (and then times out).

To complicate the situation the command prior to this has the same payload 4A 05, but has a callback and completes. I need to look more closely at the zniffer to see what is going on on the backend.  I'll do that.

If it's not repeatable on your system, it has to be the Zstick from Zooz.

My only sales pitch is to not let perfect be the enemy of good, if it is repeatable..

Bob

cdjackson:
> My only sales pitch is to not let perfect be the enemy of good, if it is repeatable..

That's fine, but if there is a timeout, reducing the time is not really the right thing to do - finding out why there's a timeout is the right approach. Otherwise you might solve this issue for you, but introducing a short timeout might have other consequences since now other commands may be sent earlier and that could cause other systems problems (eg if running on a slower computer).

apella12:
One of things I like about this is it is like peeling an onion and I'm a problem solver by heart.  Upon closer examination of the Debug files and the AddNode class code today, I realized this (add node stop command) is called twice and it's the second time with callback set to (0) that hangs.  I need to do some more testing and research to see if there is a more elegant solution.

I did look at an Zniffer extract and lined up two known commands from the Debug file above (there is a 141 msec difference between the PC time and the rpi3b time). There is no communication between the controller and the node during  either of the two add node stop commands . Picture attached
Add node stop time gap

apella12:
I closed this out of frustration with git hub.  I did solve the issue with allowing a callback on the `true` add_Node_stop.  Needed to change a test file for it to compile. may try again with guthub, but its just two lines commented out.  Successful Debug file attached
[zwave.log](https://github.com/openhab/org.openhab.binding.zwave/files/6712325/zwave.log)

Issue #1632, opened 2021-07-11
ZEN16 not reporting manual SW events due to missing COMMAND_CLASS_BASIC property


dw-8:
Since OH3 was updated in early 2021, zen16_1_0.xml had removed COMMAND_CLASS_BASIC, so OH is no longer reporting events whenever the manual switches (SW1, SW2, and SW3) are used.  Instead, the state of each switch/relay is only getting updated via polling.

Comparing OH3 with OH2.5, 
 ```
<property name="binding:*:OnOffType">COMMAND_CLASS_SWITCH_BINARY:1</property>
 <property name="binding:*:OnOffType">COMMAND_CLASS_SWITCH_BINARY:2</property> 
 <property name="binding:*:OnOffType">COMMAND_CLASS_SWITCH_BINARY:3</property>
```
should be:
```
 <property name="binding:*:OnOffType">COMMAND_CLASS_SWITCH_BINARY:1,COMMAND_CLASS_BASIC:1</property>
 <property name="binding:*:OnOffType">COMMAND_CLASS_SWITCH_BINARY:2,COMMAND_CLASS_BASIC:2</property>
 <property name="binding:*:OnOffType">COMMAND_CLASS_SWITCH_BINARY:3,COMMAND_CLASS_BASIC:3</property>

```
I made the above changes and it fixed my event reporting issue that was introduced when I upgraded from OH2.5 to OH3.0 and OH3.1.  

Also, the following eight extraneous channels were incorrectly added at the same time, and should be removed as they do not exist on the ZEN16.
```
      <channel id="switch_binary" typeId="switch_binary">
      <channel id="switch_dimmer" typeId="switch_dimmer">
      <channel id="scene_number" typeId="scene_number">
      <channel id="sensor_binary" typeId="sensor_binary">
      <channel id="meter_kwh" typeId="meter_kwh">
      <channel id="meter_kvah" typeId="meter_kvah">
      <channel id="meter_watts" typeId="meter_watts">
      <channel id="battery-level" typeId="system.battery-level">
```
The database record should also be updated to remove the extra channels:
https://opensmarthouse.org/zwavedatabase/1326

[...]

PR #1633, opened 2021-07-11
Fix Issue #1632

dw-8:
Fix issue #1632
 - change category to be consistent with ZEN16_0_0.xml (for versions <1.0)
 - remove eight extraneous channels that do not exist in ZEN16
 - add COMMAND_CLASS_BASIC property to binary switches 1-3


cdjackson:
Thanks, but please can you make the changes directly in the database. If we merge this here, it will almost immediately be reverted when the next database export is done.

https://opensmarthouse.org/blog/zwave_device_database_guide

PR #1634, opened 2021-07-16
First Time Battery node initialization- Avoid multiple rewakes

apella12:
Still struggling with Github, but I think this is complete;
Delay the go back to sleep command - Ideally until initialization is complete. Also fix the "Add Node Stop" message timeout that is reducing initial initialization time, leading to multiple wake-ups.

Attached is a composite log file of these changes on my production system with 14 battery nodes (6 different types) out of 47 total). 
[july15-zwavenode-routing.pdf](https://github.com/openhab/org.openhab.binding.zwave/files/6830933/july15-zwavenode-routing.pdf)

I modified a couple of key DEBUG logs to INFO, so that I could run the tests over several days. My focus initially was initial inclusion without a "go to sleep" command. The first section shows that with a Sensative door sensor. Although this was run with a callback number on the "Add Node Stop" message (saving 4 seconds), with the sleep timer change the node should get pretty close even with the message timeout. Compare this to a forum comment; https://community.openhab.org/t/sensative-strip-and-oh3-experience/123767/8
The second section shows the inclusion/initialization from file after OH was restarted. There are 2-3 seconds of battery savings over the default 5 second timer.

The last section shows the HEAL for each battery node. What I stumbled on was that the battery node HEAL can take longer than initial inclusion in a network of my size on distant nodes transmitting at 9.6kB (see pdf). The 18 second cap was based on earlier tests. After about 18 seconds the device can hit the Silab battery default sleep cap of 10 seconds without a command and go to sleep on its own. This cap could be shorter as for the HEAL (and Initialization for that matter) the binding will pick up where it left off. I'm running this on a pi3B, on the low end for computing speed. However, particularly with initial inclusion, I have observed some some problems on subsequent wake-ups that lead to the frequent Forum advice to keep pushing the button. I think the customer experience with Zwave would be improved with the changes in this PR. As a minor item the inclusion timer resources could be released earlier also.
[Compositetestdata.txt](https://github.com/openhab/org.openhab.binding.zwave/files/6830924/Compositetestdata.txt)


apella12:
In my effort to clear the sign-off, I've created more problems.  have to go now, but will work on later.  Anyway I think you will get the idea of what I am proposing

apella12:
Trying to clear the DCO checks pulled in a bunch of commits I had nothing to do with and I can't seem to get rid of them.  This is my second attempt at a PR and ended in failure (of a clean PR). Going to close and try again

PR #1635, opened 2021-07-16
First Time Battery node initialization- Avoid multiple rewakes

apella12:
Still struggling with Github, but I think this is complete;
Delay the go back to sleep command - Ideally until initialization is complete. Also fix the "Add Node Stop" message timeout that is reducing initial initialization time, leading to multiple wake-ups.

Attached is a composite log file of these changes on my production system with 14 battery nodes (6 different types) out of 47 total).

[july15-zwavenode-routing.pdf](https://github.com/openhab/org.openhab.binding.zwave/files/6832871/july15-zwavenode-routing.pdf)

I modified a couple of key DEBUG logs to INFO, so that I could run the tests over several days. My focus initially was initial inclusion without a "go to sleep" command. The first section shows that with a Sensative door sensor. Although this was run with a callback number on the "Add Node Stop" message (saving 4 seconds), with the sleep timer change the node should get pretty close even with the message timeout. Compare this to a forum comment; https://community.openhab.org/t/sensative-strip-and-oh3-experience/123767/8
The second section shows the inclusion/initialization from file after OH was restarted. There are 2-3 seconds of battery savings over the default 5 second timer.

The last section shows the HEAL for each battery node. What I stumbled on was that the battery node HEAL can take longer than initial inclusion in a network of my size on distant nodes transmitting at 9.6kB (see pdf). The 18 second cap was based on earlier tests. After about 18 seconds the device can hit the Silab battery default sleep cap of 10 seconds without a command and go to sleep on its own. This cap could be shorter as for the HEAL (and Initialization for that matter) the binding will pick up where it left off. I'm running this on a pi3B, on the low end for computing speed. However, particularly with initial inclusion, I have observed some some problems on subsequent wake-ups that lead to the frequent Forum advice to keep pushing the button. I think the customer experience with Zwave would be improved with the changes in this PR. As a minor item the inclusion timer resources could be released earlier also.
[Compositetestdata.txt](https://github.com/openhab/org.openhab.binding.zwave/files/6832873/Compositetestdata.txt)

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -83,9 +83,9 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
                 .withPayload(ADD_NODE_STOP).withTimeout(500).withRequiresData(false).build();
 
-//        if (complete) {
-//            payload.setCallbackId(0);
-//        }
+        if (complete) {
+            payload.setCallbackId(0);
+        }
         return payload;
     }
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -81,11 +81,11 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
 
         // Create the request
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
-                .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
+                .withPayload(ADD_NODE_STOP).withTimeout(500).withRequiresData(false).build();
 
-//        if (complete) {
-//            payload.setCallbackId(0);
-//        }
+        if (complete) {
+            payload.setCallbackId(0);
+        }
         return payload;
     }
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -83,9 +83,9 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
                 .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
 
-        if (complete) {
-            payload.setCallbackId(0);
-        }
+//        if (complete) {
+//            payload.setCallbackId(0);
+//        }
         return payload;
     }
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -83,9 +83,9 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
                 .withPayload(ADD_NODE_STOP).withTimeout(500).withRequiresData(false).build();
 
-        if (complete) {
-            payload.setCallbackId(0);
-        }
+//        if (complete) {
+//            payload.setCallbackId(0);
+//        }
         return payload;
     }
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -81,13 +81,13 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
 
         // Create the request
         if (complete) {
-            logger.debug ("Second Stop Inclusion mode -Callback 0 - Timeout")
+            logger.debug ("Second Stop Inclusion mode -Callback 0 - Timeout");
             ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
                 .withPayload(ADD_NODE_STOP).withTimeout(500).withRequiresData(false).build();
-            payload.setCallbackId(0)
+            payload.setCallbackId(0);
             return payload;
         }
-        logger.debug ("First Stop Inclusion mode Protocol")
+        logger.debug ("First Stop Inclusion mode Protocol");
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
             .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
         return payload;
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
@@ -80,12 +80,16 @@ public class AddNodeMessageClass extends ZWaveCommandProcessor {
         logger.debug("Ending INCLUSION mode.");
 
         // Create the request
-        ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
-                .withPayload(ADD_NODE_STOP).withTimeout(500).withRequiresData(false).build();
-
         if (complete) {
-            payload.setCallbackId(0);
+            logger.debug ("Second Stop Inclusion mode -Callback 0 - Timeout")
+            ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
+                .withPayload(ADD_NODE_STOP).withTimeout(500).withRequiresData(false).build();
+            payload.setCallbackId(0)
+            return payload;
         }
+        logger.debug ("First Stop Inclusion mode Protocol")
+        ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
+            .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
         return payload;
     }
 
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
@@ -1454,7 +1454,7 @@ public class ZWaveNode {
      * The timer just provides some time for anything further to be sent as a result of any processing.
      */
     private class WakeupTimerTask extends TimerTask {
-        // Two cycles through the loop are required to send a device to sleep
+        // Cycles through loop in 0.5 or 1.0 second intervals until Done or cap is reached
         private boolean triggered;
         private int count;
         private final ZWaveWakeUpCommandClass wakeUpCommandClass;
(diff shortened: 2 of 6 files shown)


cdjackson:
> Delay the go back to sleep command - Ideally until initialization is complete. 

Why? What justification is there for this? Pr why not remove the sleep timer all together?

A device does not know if the binding is in initialisation state - so it will still likely go back to sleep (unless it has been put into a mode that keeps it awake - which is rare). When the binding starts, and decides to initialise a device, this will now cause the device to be marked as DEAD.

cdjackson:
Sorry - I was being provocative :)

My point was that you have conditionally removed the sleep timer - it is removed (I think) when the binding is trying to initialse the device. But why? The device doesn't know that the binding is trying to initialise it, so it will still go to sleep if it doesn't receive commands. You will then end up with the binding marking the device as DEAD.

You also removed the 10 second timer that the binding had when the initialisation was being undertaken. Personally, I think extending the timer, not removing it is the right approach. The binding should be using a 10 second timer when the device is initialising - and removing the timer all together IMHO is a bad idea, and will result in devices being marked as DEAD.

apella12:
You definitely have more expertise and experience with what can go wrong.  My logic (and it is at least borne out on my system) is that commands will keep being fired off until initialization (or HEAL) is DONE.  Once it is DONE, there is at most one second delay before the "Go to Sleep" command is sent.  I also did cap the timer at 18 seconds. The current timer is only 5 seconds during initialization, 10 would be better for new devices and Heal, but the longer timer is also used after a restart and from my logs that only takes 2-3 seconds, even on my minimal system, so the battery use would not be optimal.

I did not run into the DEAD issue, but did find that for unlimited & 25 second cap (earlier mods I tested) the device may not be DONE and go to sleep because of the time between commands being greater than 10 seconds. You would know better, but I thought I ran across some logic that switches a battery device to asleep if a command times out, rather than DEAD.  Anyway the 18 second cap covered my slow devices on my slow system. It could be reduced, but I still like the idea of checking every second to be respectful of the battery.

cdjackson:
> My logic (and it is at least borne out on my system) is that commands will keep being fired off until initialization (or HEAL) is DONE.

But you have not changed how commands are sent - so this should happen now. I think you are effectively saying that something is delaying the commands past the 5 or 10 second delay, so the device goes to sleep. The binding is meant to send commands one ofter another now - the binding will only set the device asleep if that stops. In theory (as below) that should be after 10 seconds of no commands (the 10 seconds being 2 times 5 seconds).

> The current timer is only 5 seconds during initialization, 10 would be better for new devices and Heal

Yes, the timer is 5 seconds, but it requires 2 complete cycles before the binding sends the "Go To Sleep" command - so that is 10 seconds since the last command is sent - not 5 (or at least it should be).


>  I thought I ran across some logic that switches a battery device to asleep if a command times out, rather than DEAD

No - that is only if the "No More Information" command times out. Otherwise battery devices would never be DEAD - ever.

apella12:
> I think you are effectively saying that something is delaying the commands past the 5 or 10 second delay, so the device goes to sleep.
 I don't think I'm using that to sell these changes. However, I have measured some messages and the "find neighbors" on a battery device running 9.6 kB runs about 7 seconds and that needs to run twice during the heal, so with any additional delay the 10 seconds could be breached

> Yes, the timer is 5 seconds, but it requires 2 complete cycles before the binding sends the "Go To Sleep" command - so that is 10 seconds since the last command is sent - not 5 (or at least it should be)

It's 5 seconds total now. You split it into 2 2.5 second iterations.  All I'm doing is splitting it in up to 18 one second iterations, and if it finishes earlier it stops. 

> No - that is only if the "No More Information" command times out. Otherwise battery devices would never be DEAD - ever.
I'm not sure battery devices ARE ever dead.  The code is in the  Zwave transaction manager.  I may be mistaken but it seems if they get stuck on request they declare the node asleep.  This is under Case Cancelled
```
ZWaveNode node = controller.getNode(currentTransaction.getNodeId());
                            if (node != null && !node.isListening() && currentTransaction
                                    .getTransactionCancelledState() == TransactionState.WAIT_REQUEST) {
                                // Node is not listening, and we failed waiting for a REQUEST
                                // which means the device didn't respond. Treat as ASLEEP.
                                logger.debug("NODE {}: Transaction failed waiting for REQUEST, assume sleeping device.",
                                        currentTransaction.getNodeId());
                                node.setAwake(false);
```
My suggestions are really quite minor (and leverage the code you have already written) and I think will improve the user experience and reduce the forum traffic on battery node initialization.  All I ask is to look at my data

cdjackson:
> It's 5 seconds total now. You split it into 2 2.5 second iterations. 

Sorry, yes, you're right - 5 seconds in total for the binding to queue another message. It still seems like a hell of a lot. Maybe on slow computers it's a problem, but really 5 seconds is a long time - if every command took 5 seconds, then the initialisation would take many minutes to complete.

It's tough - we got here in the past because people complained about the battery life caused by these delays during initialisation. We had a lot of traffic about that maybe 3 or 4 (or 5?) years ago so I'm really hesitant to do something that takes us back a step here as well. I know things aren't working as well as they should at the moment, but we do need to balance this, and just increasing all the timers might not be the best approach.

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
```diff
@@ -83,9 +83,9 @@ public ZWaveSerialPayload doRequestStop(boolean complete) {
         ZWaveSerialPayload payload = new ZWaveTransactionMessageBuilder(SerialMessageClass.AddNodeToNetwork)
                 .withPayload(ADD_NODE_STOP).withRequiresData(false).build();
 
-        if (complete) {
-            payload.setCallbackId(0);
-        }
+//        if (complete) {
```
cdjackson, line 86: This is definately not good. To complete the transaction, the callback ID must be set to 0.

apella12: Without a callback the message will **always** time out on Wait_Request based on the serialmessage AddNode definitions. The only thing that I can find that needs the Callback (0) are the "TEST" files that I modified to eliminate that need.  The controller does not need it because this message is sent twice, once with a callback at the end of IncludeProtocolDone and the other at IncludeDone without the callback. I previously proposed setting a timeout on the message, like I did with the "remove node stop message", but you suggested to find the root cause which is resolved with allowing a callback ID.  Here is the debug file with a callback
[AddnodeSuccesswithcallback.txt](https://github.com/openhab/org.openhab.binding.zwave/files/6833878/AddnodeSuccesswithcallback.txt)

Also don't know if you saw this, but someone in the forum posted a partial initialization file that shows the Wait_Request cancellation on this message.  https://community.openhab.org/t/aeotec-zw130-wall-switch-not-saving-configuration/121334/3
I'm not saying this caused any of their problems, just that on an OH3 system totally unrelated to mine this command times out also.
https://community.openhab.org/t/aeotec-zw130-wall-switch-not-saving-configuration/121334/3

If you think there is some useful purpose for a Callback(0) and for the message to timeout, My alternative would be to give it a second, rather than 5.  That way it will not shorten the initialization time by about 4 seconds.  Ironically my variable sleep timer PR with a 18 second cap could handle the 4 second delay as long as another command gets sent in 10 seconds

apella12:
As to the battery life concern, the 18 seconds is a cap.  If on a fast system the initialization is done in 4 seconds, the sleep command will be sent right away.  Also I think a once-through initialization will not be any harder on batteries than multiple rewakes.  It seems to me the total awake time for device initialization should be a constant, so if it needs 9 seconds, that could be all at once (like in this PR) or over 2 x 5 second wakes.  Without going down a rabbit hole, I think there is a bit of wasted time with every wake-up to get the commands going smoothly vs getting it done in one shot. (IMO), plus the initialization timer can be cancelled earlier.

Thanks for your consideration.

Bob

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
cdjackson: > The only thing that I can find that needs the Callback (0) are the "TEST" files that I modified to eliminate that need. 

Really? Please provide the documentation reference that shows this is no longer needed. My reading of the docs clearly states that this MUST be sent to complete the transaction -:

> After receiving an ADD_NODE_STATUS_DONE status callback, the application MUST terminate the inclusion process by calling the AddNodeToNetwork function one more time. This time, the completedFunc parameter MUST be the NULL pointer.

> Being the exception to the rule, an application calling AddNodeToNetwork(ADD_NODE_STOP) to confirm the reception of a ADD_NODE_STATUS_DONE return code MUST specify the NULL pointer for the completedFunc parameter.

> If you think there is some useful purpose for a Callback(0

My understanding, from reading the docs, is that it is required - and therefore "useful". I don't see that we should change the timeout.

Maybe this has been changed in later versions of the documentation. I've not looked at the most recent version so maybe this is what you are referring to and it has changed this interface?

apella12: First my response for "useful purpose" was made in haste and I apologize as it could be construed as snarky.  It was late and I had a couple of pints, so should have turned the computer off. Again sorry, very unprofessional.

What I got when I originally read the Silabs docs was that all that is required from the controller is an "Ack" to the ADD_NODE_STOP.  Basically what you are describing more precisely.  With my very limited knowledge I was unable to modify the build message when DONE is "true" to suppress the Request encoded in the serialmessage class to handle the exception;
`.withRequiredRequest(false)` or something like it, 

So I went down the path of what is in this PR and when it worked forgot about the Silabs document.  So more accurately the combination of Callback(0) and Request (True) is the root cause of the timeout. I still think something should be done with this conflict as it is taking 4 seconds out the first 5 seconds the node is awake.

apella12 (commented):
Although I do not see that these changes will be approved, for good order I removed the non-compliance with the Silabs specification and used the timeout on the doomed message for now.  If I can figure out how to suppress the serial messages (4A & 4B) need to Wait_request on the STOP command (and therefore timeout as the callback is null, so no Request is forthcoming), I'll update the commit.
I still maintain these changes will improve the user experience, but I'm happy with them on my system.  It's quite a racehorse on initialization despite being on a rpi3B.

Bob

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
cdjackson: I'm not disagreeing that this needs to be resolved, but I don't recall this being a problem in the past so I'm a bit surprised it's a problem now since nothing has changed for a long time. I'm very hesitant to simply remove this call as it states it is required in the Silabs docs, and I have a suspicion that is it likely a security issue if inclusion is not disabled.

Really we shouldn't have this issue included here anyway as this issue is about first time battery initialisation and a PR should not include multiple features as it makes things difficult to resolve. Please split this into a separate issue.

apella12: So I'm probably not doing this correctly, but I changed this file and the other two related "test" files on the main thread yesterday because I did not want the Silabs concern to be an issue.  I reverted to the timeout option (that you also don't like), but at least there is no Silabs compliancy issue.  It is related to the topic because the message will timeout as Request(true) and Callback(0) can't coexist as message parameters.  At the default 5 seconds, the message eats up about 4 seconds of the initial initialization.   I've attached a notated extract with the baseline binding. [extractcurrentbindingSTOPcommandtimeout.txt](https://github.com/openhab/org.openhab.binding.zwave/files/6840832/extractcurrentbindingSTOPcommandtimeout.txt)

I've provided the attached file in it's full form in the forum earlier, but think there should be at least a partial here in the PR for completeness.

I had thought of three options;
1) Comment out Callback(0) - Original idea, but not Silabs Compliant- Removed
2) Shorten timeout of the message - Current edits
3) Change second "STOP" message to override the serial message Request(true), so only a low level Ack completes the command - Beyond my ability (I've tried but just get errors)

I could revert the RemoveNode file if you want as it is off topic, but is has the same issue (callback(0) & Request(True)).

I guess I need to figure out how to resolve this subthread.

Bob

apella12 (commented):
Main change today was to create two add node stop messages.  The first is like the original, the second has the timeout when the callback is 0, since that will fail anyway and is affecting the initialization time. Minor change was to change a comment to reflect the prior changes in the Node Sleep Timer and modify a log command that I am using to monitor how these changes are performing.

Still can not seem to avoid picking up a commit I had nothing to do with.

Anyway no expectations, just trying to match here what I have running
[Sleepmessagewith count.txt](https://github.com/openhab/org.openhab.binding.zwave/files/6909894/Sleepmessagewith.count.txt)

apella12:
Closing this as I have created three separate new PR

PR #1645, opened 2021-08-07
Warn if Z-wave battery device initialization or heal is not complete

apella12:
Based on a couple of recent community posting, I think this might help newer customers know, without going into Debug, that their battery device is not fully configured

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
@@ -1486,7 +1486,10 @@ public class ZWaveNode {
             }
 
             // Tell the device to go back to sleep.
-            logger.debug("NODE {}: No more messages, go back to sleep", getNodeId());
+            if (isInitializationComplete() == true) {
+                logger.debug("NODE {}: No More Messages, Go back to sleep", getNodeId());
+                }
+                logger.warn("NODE {}: Initialization or Heal not complete, but sleep anyway, state {}", getNodeId(), getNodeInitStage());
             if (wakeUpCommandClass != null) {
                 ZWaveTransactionResponse response = sendTransaction(wakeUpCommandClass.getNoMoreInformationMessage(),
                         0);


src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
```diff
@@ -1486,7 +1486,10 @@ public void run() {
             }
 
             // Tell the device to go back to sleep.
-            logger.debug("NODE {}: No more messages, go back to sleep", getNodeId());
+            if (isInitializationComplete() == true) {
+                logger.debug("NODE {}: No More Messages, Go back to sleep", getNodeId());
+                }
+                logger.warn("NODE {}: Initialization or Heal not complete, but sleep anyway, state {}", getNodeId(), getNodeInitStage());
```
nza-dk, line 1492: Aren't you missing an `else` statement, or should it always log a warning?

cdjackson:
I don't think that putting such warnings in the log is appropriate. WARN level should be for major issues - not just when someone doesn't want to use debug.

If a user wants to know the initialisation level of their device this can already be seen in the device state as this will show the level.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
apella12: So I'm not an expert on the subject, but I'm running this and the WARN only appears if "false".  It might be an allowed java shortcut with a boolean IF argument?  I could add an "else", if this is approved to merge.

apella12:
It **is** because for newer users the Debug is intimating and too much information.  Also the device state is not always shown.  I do not have a picture, but there are too many forum postings where the device is not fully configured and I find it doubtful the posters have not looked on the UI thing page.  Also I have not seen the device state updated if a heal does not complete (find neighbors, etc.)

Lastly I admire your sparse logging compared to other OH binding/core, but what about an INFO?  That's what I'm using on my set-up.

cdjackson:
Increasing the log level from DEBUG to WARN is not a solution though. This will just annoy other people as the logging will now be increased.

WARN logging is used as follows -:

* to inform the user that something seems to be wrong in the overall setup, but the system can nonetheless function as normal,
* in recoverable situations when a section of code that is not accessed under normal operating conditions is reached.

In this case it is easy to tell what the initialisation state is - just check the thing state and it will tell you. There is no need for anyone (new users or old users) to resort to looking at the debug logging (or in fact ANY logging) just to work out what the initialisation state is.

INFO logging is also not suitable -:

* INFO logging should be used sparsely. e.g. a newly started component or a user file that has been loaded.

Again, this is not the right approach - just look at the thing state if you want to know what the state is.

apella12:
1) Ok.  Fine
2) I still maintain it is not always shown on the thing state, but since I'm using the variable timer, I don't have the problem situation, so can't show picture. When I get one, I'll update this
3) as to INFO I have totally suppressed the events log, but this is the default. (see file
[events.log](https://github.com/openhab/org.openhab.binding.zwave/files/7003787/events.log)
)  I really don't think one more INFO is going to matter, but it's your binding.

Bob

cdjackson:
> I still maintain it is not always shown on the thing state

If that is the case, then that should be resolved rather than adding warnings.

> ) I really don't think one more INFO is going to matter, but it's your binding.

What I posted above is the guidelines on what to log. I get a lot of complaints already about too much logging. The logging guidelines state that INFO should be used sparsely. If someone has a number of devices, then it's not just one more entry - it will be dozens or more per day.

cdjackson:
> No one in their right mind can complain about excess zwave logging.

And yet they do - not so much now as I got pushed to lower a lot of logging, but just search through the forum and you will see plenty of complaints about excessive logging.

> I'd also add that by logging a user can see if there was a problem historically (like overnight)

No - this doesn't log errors - this is just logging routine information. It will not tell you if there was a problem overnight. The initialisation is routine - it should progress sequentially, so when the user wakes up in the morning they can look at the state - if the initialisation is complete, then any logging overnight is irrelevant. If it's stuck, then seeing that it didn't progress over night doesn't help at all. The only thing that is important is the current state. Otherwise just enable debug logging - it's not that hard!

I'll close this one but please feel free to look at the option of ensuring the state is updated correctly.

PR #1653, opened 2021-08-16
[zwave] Execute slow dispose() actions on a scheduler thread

andrewfg:
**Background:**
The `ZWaveControllerHandler.dispose()` method violates the openHAB coding guidelines because it takes a very long time (ca. 6 seconds) to complete its actions (see screenshot below).

This delay can cause problems in shutting down other bindings because it starves them of sufficient time to shut themselves down properly [e.g. Velux binding](https://community.openhab.org/t/velux-new-openhab2-binding-feedback-welcome/32926/1311)

image

**Solution:**
This PR shifts the long running `ZWaveControllerHandler.dispose()` tasks to a separate thread run on the scheduler. So the `dispose()` can itself return immediately.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
@@ -191,7 +191,6 @@ public abstract class ZWaveControllerHandler extends BaseBridgeHandler implement
         }
 
         initializeHeal();
-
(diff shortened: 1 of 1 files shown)


cdjackson:
Thanks. I'm not sure why this takes so long - it should be quick I would have thought, but in any case this is fine.

andrewfg:
Many thanks @cdjackson

PR #1657, opened 2021-08-31
Add conversion for differential white report

Aaargh20318:
On devices with a differential white channel (warm+cold white), such as the Zipato Bulb 2, the value of the color temperature channel always resets to 0 after updating. While the color temperature value is converted to a warm + cold value when setting the value, the reverse conversion isn't performed when receiving the report back from the device. 

This PR adds the reverse conversion so the channel is updated correctly.


cdjackson (changes requested):
This looks fine, but please format the code using the formatter.

Aaargh20318:
I ran the formatter and it changed nearly every file for some reason. Makes the patch a bit messy.

cdjackson:
Are you sure you are using the correct formatter for OH? Either that or manually format it - at the moment it's really not good as all the indentation is messed up.

Aaargh20318:
> Are you sure you are using the correct formatter for OH? Either that or manually format it - at the moment it's really not good as all the indentation is messed up.

I ran `mvn spotless:apply`, as that's what's configured in the POM. 

Should I be using a different formatter ?

cdjackson:
I normally use the Eclipse IDE formatter. In any case it needs resolving before merging please.

Aaargh20318:
I'll revert the formatting changes and just do it manually.

cdjackson:
Thanks. This looks fine now.

cdjackson (approved):

Issue #1670, opened 2021-10-14
ZCOMBO smoke alarm does not intialize


marcusb:
* OH2 Version: 3.1.0
* Binding Version (if SNAPSHOT, provide compile time as found on the filename in the console): 3.1.0

If the problem is associated with a device, please provide the following -:

* Device name: ZCOMBO Smoke Detector
* Link to [database](http://opensmarthouse.org/zwavedatabase): https://opensmarthouse.org/zwavedatabase/238

This device never initializes, despite dozens of manual wakeup attempts using the button. Each time it gets stuck in the same state and fails to make progress.

```
2021-09-11 20:17:26.215 [DEBUG] [essage.ApplicationUpdateMessageClass] - NODE 24: Application update request. Node information received. Transaction null
2021-09-11 20:17:26.217 [DEBUG] [essage.ApplicationUpdateMessageClass] - NODE 24: Application update - no transaction.
2021-09-11 20:17:26.468 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 24: Is awake with 1 messages in the queue
2021-09-11 20:17:26.469 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 24: COMMAND_CLASS_WAKE_UP not found - setting AWAKE
2021-09-11 20:17:26.471 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 24: Start sleep timer at 5000ms
2021-09-11 20:17:26.471 [DEBUG] [ding.zwave.handler.ZWaveThingHandler] - NODE 24: Got an event from Z-Wave network: ZWaveNodeStatusEvent
2021-09-11 20:17:26.475 [DEBUG] [ve.internal.protocol.ZWaveController] - NODE 24: Node Status event - Node is AWAKE
2021-09-11 20:17:28.972 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 24: WakeupTimerTask 1 Messages waiting, state MANUFACTURER
2021-09-11 20:17:31.471 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 24: WakeupTimerTask 1 Messages waiting, state MANUFACTURER
2021-09-11 20:17:31.472 [DEBUG] [ng.zwave.internal.protocol.ZWaveNode] - NODE 24: No more messages, go back to sleep
```

The issue has been reported in the forum by several users.

PR #1671, opened 2021-10-14
Increase sleep timer delay when initializing a device (#1670)

marcusb:
Some devices are slow to respond during intialization, and a timeout
of 5000 ms is not enough to let them progress. Examples include the
ZCOMBO smoke alarm.

Increasing the timeout interval to 30 seconds allows the smoke alarm
to finish initialization within a few wakeups.

Signed-off-by: Marcus Better <marcus@better.se>


cdjackson:
> Some devices are slow to respond during intialization, and a timeout

Please note that this timer is not about how long it takes the device to respond - it is about how long it takes the BINDING to respond.

Without knowing the particular device it's really hard to comment on why it might not be sufficient.

marcusb:
Closing in favor of https://github.com/openhab/org.openhab.binding.zwave/pull/1647

PR #1681, opened 2021-11-08
fix flaky test in Alarm_AlarmRaw

YujiaYan0709:
For the `Alarm_AlarmRaw` test in `ZWaveAlarmConverterTest.java`, since it used hashmap to return the State object, the order of hashmap key-value pair is non-deterministic. Therefore, directly converting to a String and comparing is a flaky test, and I fix it by comparing each key-value pair.


cdjackson:
While technically you're correct, I've never seen this test fail. Anyway, I'm happy to make it more robust, but please remove all the binary files you are adding here first.

YujiaYan0709:
I have removed all ds files. Sorry for the inconvenience. Let me know if there is any other issue.

PR #1689, opened 2021-11-26
Fix build

wborn:
The build fails because the code uses deprecated APIs which were removed in Mockito 4.

Related to: https://github.com/openhab/openhab-core/pull/2574

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/initialization/ZWaveNodeSerializer.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/initialization/ZWaveNodeSerializer.java
@@ -40,7 +40,7 @@ import com.thoughtworks.xstream.io.xml.StaxDriver;
 /**
  * ZWaveNodeSerializer class. Serializes nodes to XML and back again.
  *
- * @author Chris Jackson
+ * @author Chris Jackson - Initial contribution
  * @author Jan-Willem Spuij
  */
 public class ZWaveNodeSerializer {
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Trying to rebuild.

wborn:
It's very stubborn! You'd almost think it's a real issue. :wink:

nkasrawi:
I hope this absolves my PR!

wborn:
It seems to build correctly now @openhab/z-wave-maintainers + @kaikreuzer !

wborn:
It does seem to fail locally now, which can be fixed by adding `@Execution(ExecutionMode.SAME_THREAD)` again.
Let's see how that works out on Jenkins.

apella12:
I build a local zwave .jar for my own testing and use from the code on this page (with my mods of course).  I ran into these compile errors when I made some recent code changes on my download from November 1, 2021.  I did pick up on the first file change and then ran into the serialize/deserialize error and decided to give-up until I saw it was fixed (I have little knowledge compared to you all).  I only mention this because my code was from earlier, but the error messages are the same.  Usually this means something changed in the files downloaded for the build (like when Xstream was recently upgraded and one line was removed from the zwave binding). Also it is not the recent PR by @nkasrawi because I do not have that.

kaikreuzer:
Hooray, a green build - thanks @wborn!

kaikreuzer (approved):

wborn:
It's probably better to not use this `ExecutionMode`, because it is an experimental feature:

```java
/**
 * Supported execution modes for parallel test execution.
 *
 * @since 5.3
 * @see #SAME_THREAD
 * @see #CONCURRENT
 */
@API(status = EXPERIMENTAL, since = "5.3")
public enum ExecutionMode {
```

```java
		/**
		 * Intended for new, experimental features where the publisher of the
		 * API is looking for feedback.
		 *
		 * <p>Use with caution. Might be promoted to {@link #MAINTAINED} or
		 * {@link #STABLE} in the future, but might also be removed without
		 * prior notice.
		 */
		EXPERIMENTAL,
```

PR #1694, opened 2021-12-10
Delay sleep command until device is initialized

apella12:
Don't mind me.  Trying to change name and add a missing "else" statement


apella12:
Tried to correct a mistake and it picked up a bunch of commits

PR #1695, opened 2021-12-10
Delay Sleep Command until battery device message queue is empty

apella12:
Modified the timer to use the device queue as trigger. Still need a timer cap on the iterations in case device gets stuck.  Did some custom logging using the info level (since there is very little in the zwave binding at the info log level, I can use Log:set info & then  log:set Warn when I get finished testing without having to change the binding.  I'm testing on my production system, so don't like to change the binding.
Anyway, my observation is that it produces about the same result as using initialization as the trigger, but technically covers the case when the device is DONE, but there is over 1 second of messages piled up.
[annotatedzwavelog.txt](https://github.com/openhab/org.openhab.binding.zwave/files/7694425/annotatedzwavelog.txt)
)

No need to comment, I did as an academic exercise


apella12:
closing with One and DONE package

PR #1696, opened 2021-12-10
Delay Sleep Message until Battery device is Initialized

apella12:
This is a rehash to fix an error and Github problem.  No need to comment and plan on closing.  Just wanted a clean version


apella12:
Sleep message keys off initialization rather than battery device queue.  I used for several months, but have switched over to the device queue trigger for now.

PR #1703, opened 2021-12-20
Adding channel type for raw control of color

jjlauterbach:
Adding a channel type to support raw control of a color light.  This follows up #1675 

Signed-off-by: Jeff Lauterbach <jjlauterbach@gmail.com>

### DIFF
--- a/src/main/resources/OH-INF/thing/channels.xml
+++ b/src/main/resources/OH-INF/thing/channels.xml
@@ -414,6 +414,14 @@
         <category>ColorLight</category>
     </channel-type>
 
+    <!-- Color Command Class - Raw -->
+    <channel-type id="color_raw">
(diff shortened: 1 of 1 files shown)


jjlauterbach:
@cdjackson Will this new channel type automatically roll out to the zwave database?  Or is there another repo that needs to be updated as well?

cdjackson:
It’s not automatic, but it should be already in the list as I added it shortly after you created this PR.

jjlauterbach:
Not seeing it yet in the ui

cdjackson:
I just had a look, and it is there. I did note though that the list is not sorted, so it’s right at the bottom.

jjlauterbach:
Okay I cleared browser cache for the site and now its showing up.  Thanks

PR #1715, opened 2022-01-08
Add support for COMMAND_CLASS_SOUND_SWITCH

### LINKED ISSUE #1245: Add support for COMMAND_CLASS_SOUND_SWITCH
This device needs significant binding updates.

"there are significant code changes required to support this device as it uses a new command class."
from https://community.openhab.org/t/aeotech-doorbell-6/79278/42

Also, since the database maintenance, this device is no longer visible in the summary.

kennetn:
Adds new COMMAND_CLASS_SOUND_SWITCH needed to be able to trigger siren and volume on Aeotec Siren/Doorbell 6.
This pull request is more polished than the previous one and I am not aware of any bugs. I also tried to keep in line with the rest of the code.

The ZWaveSoundSwitchCommandClass.java is strait forward and made like all the other command classes.
But because the command class supports getting and setting status of both volume and the tone to play / being played.
I needed to differentiate between the two in the ZWaveSoundSwitchConverter.java. I could not find any other classes that did this so I was a bit unsure how I should do this.
I decided to do this with `channel.getChannelTypeUID().getId().equals("sound_volume")` I guess I could also have used parameters or maybe multiply Converter classes. But I think there is always a 1 to 1 relatiationship between command class and converter. So I thought this was the right way to do it.
I therefore added two new channel types (sound_tone_play and sound_volume) in channels.xml but again I was not sure how they should be named. If there is some schema in the names that I did not get, let me know and I will update.

I guess the two new types need to be added to the Z-wave database editor also, is there any way I can help with that ?

The tone_play channel will be updated when the tone has finished playing.

Closes #1245
Signed-off-by: kennetn <kennetn@gmail.com>

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveCommandClassConverter.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/converter/ZWaveCommandClassConverter.java
@@ -69,6 +69,7 @@ public abstract class ZWaveCommandClassConverter {
         temp.put(CommandClass.COMMAND_CLASS_SENSOR_ALARM, ZWaveAlarmSensorConverter.class);
         temp.put(CommandClass.COMMAND_CLASS_SENSOR_BINARY, ZWaveBinarySensorConverter.class);
         temp.put(CommandClass.COMMAND_CLASS_SENSOR_MULTILEVEL, ZWaveMultiLevelSensorConverter.class);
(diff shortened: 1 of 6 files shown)


cdjackson:
@kennetn sorry for the very slow response.

Please can you rebase this on the current master and update license headers to the 2022.

I also need to update the database to add these channels...

kennetn:
Hi @cdjackson 
Thanks for looking at this. I did a rebase.

cdjackson:
Thanks @kennetn - looks good.

PR #1725, opened 2022-01-29
Add instructions for excluding Z-Wave devices

poisonsnak:
Currently the instructions say that if you want to exclude a device, you should put the controller into include mode.  This doesn't work.  You have to put it into exclude mode by going to the controller Thing itself and selecting Exclude Devices.  I just tried it myself and confirmed that include mode doesn't work but exclude mode does.  Chris confirms that this is the correct procedure here as well: https://community.openhab.org/t/how-to-exclude-a-z-wave-device-with-openhab-3/120127/6.

Signed-off-by: Jared Epp <jaredmepp@gmail.com> (github: poisonsnak)


cdjackson:
Thanks @poisonsnak

PR #1732, opened 2022-02-07
Update ZWaveMultiLevelSensorCommandClass.java

apella12:
Forum issue https://community.openhab.org/t/trane-xr524-zwave-not-updating-oh3/114703/18

I'm a little out of my comfort zone, don't have the device, but the forum poster (last several posts) has a lot of information and I thought I would take a shot.  He modified to get degrees F, but by my novice java code reading this class should do this automatically, if the conditional (IF) is changed not to overwrite.  probably wrong but wanted you to have a look.
Bob


apella12:
Well I took a SWAG based on the posting and it was wrong.

PR #1736, opened 2022-02-18
“One and DONE” - Zwave battery device optimization

apella12:
This PR is to optimize the user experience of initializing new Zwave battery devices. The desired result is to be able to put a battery in the shiny new device, follow the device manual inclusion protocol during an Inbox, Zwave binding “Scan” mode of OpenHab and, in 20 seconds or less, have all the channels displayed and begin to link them to items.  “One and DONE”.
Since my last unmerged battery device PRs I have replaced the fixed 20 second cap in the iterative sleep timer with a user defined maximum “hold awake” parameter.  As the user defined default is five seconds, there is no implied maintainer approval of a longer awake duration.  Kind of a “use at your own risk” feature.  Once the device is DONE (or the cap is reached) the sleep message (No More Information-NMI) is created.  Setting a higher user defined cap does not mean increased battery usage.  With half second intervals, versus 5 seconds in today’s binding, daily battery device awake duration is either the same or reduced.
Also added is a kick-Queue action after the half second mark in the Timer Task to get a potentially “stuck” device queue going again. In some scenarios (primarily during initialization, the existing kick-Queue action appears to happen too early (before the event listener sets the device awake). There are examples from the forum in my attached document.  I was able to recreate the “stuck” scenario (by reverting the second “Add Node Stop” & Sleep Timer cap to 5 seconds) and resolve the “stuck” problem on my test device with this additional kick-Queue.
Lastly, I have added the shortened timeout for the second calling of the “Add Node Stop” to make this a complete package.  It has the most impact as anything else to “One and DONE”.  The default 5 seconds to cancel this command eats up about 4 seconds of the Timer leaving only one second left in the initial awake interval. Some of the other enhancements in this PR would have less impact, if a full 5 seconds were available for device initialization after device discovery.
I have all the changes in this PR running, some for as long as 8 months (as of Jan 2022). I have 13 battery nodes of various manufacturers (47 total) on an Rpi4 with an Rpi3b test system. It is frustrating to see the battery device initialization issues in the forum, as I know they can be resolved better than the standard “reawake many, many times” advice. These changes will save considerable user frustration and many seconds of battery usage.


apella12:
While trying to understand github better I stumbled upon where I could see the history of the Zwave sleep timer in the ZwaveNode class. Most enlightening in regards to this PR was #1100.  Here you removed the check for messages in the queue because, as sometimes happens in Zwave communications, the message queue is not always cleared and the device was still assumed to be awake by the binding. This PR puts that check back (together with the Node status), but with the maximum wake duration cap structure overlaid the issue in #1100 will not occur. After the cap is reached the sleep message will be sent regardless of messages or node status.
Also worth a comment is PR #1134. Here you adjusted the sleep timer duration due to initiation problems like this PR addresses.  All my PR does differently is use the functionality of the java timer schedule to check the message queue and Node state at 0.5 second intervals, rather than guess if 2.5 or 5.0 seconds is the right level for all OH configurations and network sizes.
Lastly is PR #1238 to add a KickQueue due to a 250ms delay in some circumstances.  From my log file contained in this PR (justifying an additional KickQueue) the current KickQueue for a device that has not had the WakeUp CC initialized is still 250ms too early. This PR sends a KickQueue at the 0.5 second mark and gets initiation restarted.
Testing Update:
I have attached a custom log file from my system for one month with this PR running (some tweaks here and there).  What I have done, since the Zwave INFO level is used sparingly and my devices only wake daily is to adjust certain DEBUG levels in the Sleep Timer section of the ZwaveNode class to INFO, so I can get several weeks of data in a smallish file.  In the actual PR code all but one log item is at Debug. There is a mix of initiations, heals, restarts and normal wakes (when status is DONE) in decreasing frequency. Rarely is the cap reached. Average awake time with a 19 second cap is 1.47 seconds.
PR Test-1month
[One and DonePR.txt](https://github.com/openhab/org.openhab.binding.zwave/files/8287190/One.and.DonePR.txt)
Bob

MyRaceData:
@cdjackson wdyt?

cdjackson:
> @cdjackson wdyt?

Sorry - I've been very busy over the past couple of months so will try and look at this shortly

mhilbush:
Chris, I realize you're busy, but have you had a chance to look at this? It would be very nice to have a more efficient inclusion process for battery powered devices. Some of these devices have quite a few channels and config parameters, resulting in the need to do many wake ups in order to complete the initialization. In some recent cases, I've had to wake up the device nearly a dozen times to complete initialization.

apella12:
So in changing the timing of the kickQueue from 0.5 seconds to 2 seconds, I completely messed this up again. I do not know why it adds me to commits I had nothing to do with. I hate github. If this is going to be merged I'm going to have to start over.  The only two commits that need review are 9e9c7fd and 08a4546

cdjackson:
The problem is you are using a merge commit - you should always rebase on master, not merge master into your branch.  This way, your changes get added to the end of current head of master and not the other way around.

Sorry - please can you resolve this and I will try and take a look over the weekend.

apella12:
I'll do it tomorrow.  Sorry. 

Bob

PR #1760, opened 2022-04-27
Zwave battery device optimization

apella12:
This PR is to optimize the user experience of initializing new Zwave battery devices. The desired result is to be able to put a battery in the shiny new device, follow the device manual inclusion protocol during an Inbox, Zwave binding “Scan” mode of OpenHab and, in 20 seconds or less, have all the channels displayed and begin to link them to items. “One and DONE”.
Since my last unmerged battery device PRs I have replaced the fixed 20 second cap in the iterative sleep timer with a user defined maximum “hold awake” parameter. As the user defined default is five seconds, there is no implied maintainer approval of a longer awake duration. Kind of a “use at your own risk” feature. Once the device is DONE (or the cap is reached) the sleep message (No More Information-NMI) is created. Setting a higher user defined cap does not mean increased battery usage. With half second intervals, versus 5 seconds in today’s binding, daily battery device awake duration is either the same or reduced.
Also added is a kick-Queue action after the half second mark in the Timer Task to get a potentially “stuck” device queue going again. In some scenarios (primarily during initialization, the existing kick-Queue action appears to happen too early (before the event listener sets the device awake). There are examples from the forum in my attached document. I was able to recreate the “stuck” scenario (by reverting the second “Add Node Stop” & Sleep Timer cap to 5 seconds) and resolve the “stuck” problem on my test device with this additional kick-Queue.
Lastly, I have added the shortened timeout for the second calling of the “Add Node Stop” to make this a complete package. It has the most impact as anything else to “One and DONE”. The default 5 seconds to cancel this command eats up about 4 seconds of the Timer leaving only one second left in the initial awake interval. Some of the other enhancements in this PR would have less impact, if a full 5 seconds were available for device initialization after device discovery.
I have all the changes in this PR running, some for as long as 8 months (as of Jan 2022). I have 13 battery nodes of various manufacturers (47 total) on an Rpi4 with an Rpi3b test system. It is frustrating to see the battery device initialization issues in the forum, as I know they can be resolved better than the standard “reawake many, many times” advice. These changes will save considerable user frustration and many seconds of battery usage.


apella12:
This replaces 1736 (and the other 11 closed PRs in my zwave battery device journey to help the OH community and get github to cooperate with my thought processes).  Last night I considered dropping this.  It works flawlessly for me and I know from prior comments you don't want me messing with your code and there is no appeal process, but in the end that seemed selfish and not in the spirit of open source and the help so many have provided to get me where I am today with home automation.  Below are my additional comments from #1736. You really need to check PR #1100, if nothing else, as a statement you made on a previous PR was not correct in regards to whether the sleep message is delayed if messages are in the queue.  It was true, but not now, but will be again with this PR.

Bob
-------------
While trying to understand github better I stumbled upon where I could see the history of the Zwave sleep timer in the ZwaveNode class. Most enlightening in regards to this PR #1100 . Here you removed the check for messages in the queue because, as sometimes happens in Zwave communications, the message queue is not always cleared and the device was still assumed to be awake by the binding. This PR puts that check back (together with the Node status), but with the maximum wake duration cap structure overlaid,  the issue in #1100 will not occur. After the cap is reached the sleep message will be sent regardless of messages or node status.
Also worth a comment is PR #1134. Here you adjusted the sleep timer duration due to initiation problems like this PR addresses. All my PR does differently is use the functionality of the java timer schedule to check the message queue and Node state at 0.5 second intervals, rather than guess if 2.5 or 5.0 seconds is the right level for all OH configurations and network sizes.
Lastly is PR #1238 to add a KickQueue due to a 250ms delay in some circumstances. From my log file contained in this PR (justifying an additional KickQueue) the current KickQueue for a device that has not had the WakeUp CC initialized is still 250ms too early. This PR sends a KickQueue at the 2 second mark and gets initiation restarted, if it is hung up.

Testing Update:
I have attached a custom log file from my system for one month with this PR running (some tweaks here and there). What I have done, since the Zwave INFO level is used sparingly and my devices only wake daily is to adjust certain DEBUG levels in the Sleep Timer section of the ZwaveNode class to INFO, so I can get several weeks of data in a smallish file. In the actual PR code all but one log item is at Debug. There is a mix of initiations, heals, restarts and normal wakes (when status is DONE) in decreasing frequency. Rarely is the cap reached. Average awake time with a 19 second cap is 1.47
image

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
```diff
@@ -1475,27 +1477,41 @@ public void run() {
...
+                    controller.getSendQueueLength(getNodeId()), getNodeInitStage(), count);
 
-            logger.debug("NODE {}: WakeupTimerTask {} Messages waiting, state {}", getNodeId(),
-                    controller.getSendQueueLength(getNodeId()), getNodeInitStage());
-            if (triggered == false) {
-                logger.trace("NODE {}: WakeupTimerTask First iteration", getNodeId());
+            if (isInitializationComplete() == true  && controller.getSendQueueLength(getNodeId()) == 0) {
+                triggered = true;
+            }     
+            if (count == awakemax || count == 40) {
```
cdjackson, line 1487: Why count = 40?
Constants should be defined in the top of the class.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
```diff
@@ -1475,27 +1477,41 @@ public void run() {
...
+                logger.trace("NODE {}: WakeupTimerTask iteration", getNodeId());
                 return;
             }
 
-            // Tell the device to go back to sleep.
-            logger.debug("NODE {}: No more messages, go back to sleep", getNodeId());
+            // Tell the device to go back to sleep. This message may be queued if the prior message timed out
+            // and the node was therefore assumed to be asleep.  It can disrupt a future awake.
+            if (isInitializationComplete() == true  && controller.getSendQueueLength(getNodeId()) == 0) {
+            logger.debug("NODE {}: Go back to sleep, state {}, count {} messages {}", getNodeId(), getNodeInitStage(), count, controller.getSendQueueLength(getNodeId()));
```
cdjackson, line 1501: Formatting  / indentation seems incorrect around here.

src/main/resources/OH-INF/thing/controller_serial.xml
```diff
@@ -72,6 +72,13 @@
                 <advanced>true</advanced>
             </parameter>
 
+            <parameter name="controller_maxbatterywakeduration" type="integer" groupName="network" min="5" max="20">
```
cdjackson, line 75: I'd suggest to make this less that 5 seconds - many people like it to be a lot less (this was a very hot topic a few years ago which is why I had to trim it down). I'd suggest 2 seconds.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java
```diff
@@ -1087,6 +1092,10 @@ public int getSystemDefaultWakeupPeriod() {
         return defaultWakeupPeriod;
     }
 
+    public int getSystemMaxBatteryWakeDuration() {
```
cdjackson, line 1095: Please add java doc here.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
```diff
@@ -1475,27 +1477,41 @@ public void run() {
...
-            if (triggered == false) {
-                logger.trace("NODE {}: WakeupTimerTask First iteration", getNodeId());
+            if (isInitializationComplete() == true  && controller.getSendQueueLength(getNodeId()) == 0) {
+                triggered = true;
+            }     
+            if (count == awakemax || count == 40) {
                 triggered = true;
+            }
+            if (triggered == false) {
+                if (count == 4) {
```
cdjackson, line 1491: Why count=4?

constants should be defined at the top of the class.

cdjackson (changes requested):
Thanks @apella12 and sorry for the very long delay in looking at this.  I think with the check that initialisation is not complete (which I'd forgotten had been removed) I'm happy with this approach. My only lingering concern is that effectively you are largely restoring a change that we previously removed, and presumably I removed it for a reason, so that might bite us again at some point.

Otherwise most of the issues here are just formatting and naming.

apella12:
Thanks for your review. Just to let you know I am working on your changes. 

This is all new to me, so will need to do some Github research, not so much on the "what"- (I get most of your requests), but the "how" to do this in a way not to mess up this PR and have to start over.  I think the proper approach is to make my changes locally and then amend (or rebase) a commit to this PR. If you have any advice on the "how" from your experience, I'd like to know.

As to your overall concern, the stated reason in the PR #1100 to remove the message queue check was if the message queue was not cleared, the device would never be set as "asleep" by the binding, messages would still be sent, and the device would be marked as dead.  I'm optimistic adding the maxAwakePeriod (your name suggestion is fine) as an override should solve that problem in most, if not all the cases.  Over my testing period, (nearly 6 months) I have not had a problem with battery devices being marked "Dead".

The only lingering concern I have that we are not fully on the same page was your comment about lowering the minimum maxAwakePeriod to 2 seconds in the controller_serial.xml file. The maxAwakePeriod is meant as an override and only come into play about 1% (or less) of the time (with a message stuck in the queue).  At two seconds the cap will interfere with battery device initialization and heal. My first motivation for making this a user defined parameter was your prior comments regarding 5 seconds.  This way I could notionally keep the 5 seconds but stand ready to recommend to any user to raise this to at least 10 seconds. Note my chart above with the maxAwakePeriod set at 19 seconds. I have tried to show there is not a battery usage issue, because once all the messages are sent and the device initialized, the device is put back to sleep.   Most "awakes" (say just for a battery check) take 0.5 seconds. (OH restarts require 1-2 seconds, Initializations take 3-8 seconds, and "Heals" take 10-15 seconds). I expect these times to vary depending the size of the network and the number of parameters available in the battery devices.  My battery devices have held their charge over the last several months.

Will get to work on these changes and thanks again for your review.

Bob

mhilbush:
> This is all new to me, so will need to do some Github research, not so much on the "what"- (I get most of your requests), but the "how" to do this in a way not to mess up this PR and have to start over. I think the proper approach is to make my changes locally and then amend (or rebase) a commit to this PR. If you have any advice on the "how" from your experience, I'd like to know.

All you should need to do is make the changes to your local development branch. Then when you're happy with the changes, commit those changes (don't forget to sign your commit), then push the branch. There should be no need to do a merge or rebase.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
mhilbush: I'm also a bit confused about what you're trying to accomplish here.

It looks like you're allowing this timer task to execute up to `awakemax` or 40 times (based on line 1487), but you're only kicking the queue on iteration #4. Why?

Or, were you intending to kick the queue on every 4th iteration of the timer task? If so, checking for `count == 4` isn't going to do that.

cdjackson:
> the "how" to do this in a way not to mess up this PR and have to start over

As Mark said - just make the changes, commit them into the existing branch, and push them to GitHub.

> At two seconds the cap will interfere with battery device initialization and heal.

I'm not sure why - with the check to see if there are still frames queued, then it should be fine?

> the stated reason in the PR https://github.com/openhab/org.openhab.binding.zwave/pull/1100 to remove the message queue check was if the message queue was not cleared, the device would never be set as "asleep" by the binding

Ok, it's coming back now :) I'm happy that your max counter addresses this issue.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
cdjackson: Looking at this some more, I don't understand why you have the `|| count == 40` part at all. This seems to be another catchall in case `awakemax` is set to some random number, but that's not possible as this configuration is already limited to 20 seconds.

I would suggest to remove this check as it effectively conflicts with the configuration setting, and it means if we ever were to change the configuration, it wouldn't work due to this check.

src/main/java/org/openhab/binding/zwave/ZWaveBindingConstants.java
```diff
@@ -56,6 +56,7 @@ public class ZWaveBindingConstants {
 
     public final static String CONFIGURATION_POLLPERIOD = "binding_pollperiod";
     public final static String CONFIGURATION_CMDREPOLLPERIOD = "binding_cmdrepollperiod";
+    public final static String CONFIGURATION_MAXBATTERYWAKEDURATION = "controller_maxbatterywakeduration";
```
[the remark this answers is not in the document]
apella12: done
`public final static String CONFIGURATION_MAXAWAKEPERIOD = "controller_maxawakeperiod";`

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -118,6 +119,13 @@ public void initialize() {
             wakeupDefaultPeriod = 0;
         }
 
+        param = getConfig().get(CONFIGURATION_MAXBATTERYWAKEDURATION);
+        if (param instanceof BigDecimal) {
+            maxbatteryawake = ((BigDecimal) param).intValue();
+        } else {
+            maxbatteryawake = 5;
```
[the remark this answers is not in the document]
apella12: Unless I'm missing something, it already is there (at the top of the class-see comment below).  I copied/patterned this off
`private Integer wakeupDefaultPeriod;`  from the controller_serial.xml, to the ZWaveBindingConstants, to the ZWaveControllerHandler, to the ZWaveController

This section is now;
```
        param = getConfig().get(CONFIGURATION_MAXAWAKEPERIOD);
        if (param instanceof BigDecimal) {
            maxAwakePeriod = ((BigDecimal) param).intValue();
        } else {
            maxAwakePeriod = 5;
        }
```

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -69,6 +69,7 @@ public abstract class ZWaveControllerHandler extends BaseBridgeHandler implement
     private Integer secureInclusionMode;
     private Integer healTime;
     private Integer wakeupDefaultPeriod;
+    private Integer maxbatteryawake;
```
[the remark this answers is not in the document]
apella12: Done
`private Integer maxAwakePeriod;`

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -522,6 +533,10 @@ public Integer getDefaultWakeupPeriod() {
         return wakeupDefaultPeriod;
     }
 
+    public Integer getMaxBatteryWakeDuration() {
```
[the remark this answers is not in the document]
apella12: Done
```
    /**
     * Gets the default max awake period configured for this network
     *
     * @return the default wakeup, or null if not set
     */

    public Integer getMaxAwakePeriod() {
        return maxAwakePeriod;
    }
```

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java
```diff
@@ -107,6 +107,7 @@ public class ZWaveController {
 
     private ZWaveInclusionController inclusionController = null;
     private int defaultWakeupPeriod = 0;
+    private int maxbatterywakeduration = 5;
```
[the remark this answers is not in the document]
apella12: Done
`private int maxAwakePeriod = 5;`

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
```diff
@@ -1448,15 +1448,16 @@ public boolean isAwake() {
...
      * The timer just provides some time for anything further to be sent as a result of any processing.
      */
     private class WakeupTimerTask extends TimerTask {
-        // Two cycles through the loop are required to send a device to sleep
+        // Sleep message after max battery awake duration or 20 seconds
         private boolean triggered;
+        private int count;
         private final ZWaveWakeUpCommandClass wakeUpCommandClass;
-
+        private int awakemax = (controller.getSystemMaxBatteryWakeDuration() * 2);
```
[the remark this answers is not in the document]
apella12: Done
`private int awakeMax = (controller.getSystemMaxAwakePeriod() * 2);`

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
```diff
@@ -77,16 +77,20 @@ public ZWaveSerialPayload doRequestStart(boolean highPower, boolean networkWide)
     }
```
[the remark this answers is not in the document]
apella12: It is totally related because the message without a callback eats up 4 seconds of the initialization of battery devices before it is cancelled.  It is really critical to the "One and Done" idea.  I implore you to reconsider, no I beg you.  We really would need to raise the default cap to 10 seconds to counteract leaving this in.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
apella12: Done - I took this out.  It was a redundant backstop, but the awakeMax should be enough.
```
            if (count == awakeMax) {
                triggered = true;
            }
```

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
apella12: Done, (I think)
```
            // Tell the device to go back to sleep. This message may be queued if the prior message timed out
            // and the node was therefore assumed to be asleep.  It can disrupt a future awake.
            if (isInitializationComplete() == true  && controller.getSendQueueLength(getNodeId()) == 0) {
                logger.debug("NODE {}: Go back to sleep, state {}, count {} messages {}", getNodeId(),
                    getNodeInitStage(), count, controller.getSendQueueLength(getNodeId()));
            }
            else {
                logger.info("NODE {}: Wake duration reached, state {} count {}, messages {}", getNodeId(),
                    getNodeInitStage(), count, controller.getSendQueueLength(getNodeId()));
            }
```

src/main/resources/OH-INF/thing/controller_serial.xml
apella12: I lowered the minimum to 2, but it will blow up on anyone that tries it, especially if we can't keep the add Node serial message changes. The sleep message will get queued before the device can do ANY configuration
          `  <parameter name="controller_maxawakeperiod" type="integer" groupName="network" min="2" max="20">`

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java
apella12: Done
```
    /**
     * Gets the system wide default period for max awake time
     *
     * @return the awake period in seconds, or 5 if no default is set
     */
    public int getSystemMaxAwakePeriod() {
        return maxAwakePeriod;
    }
```

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
apella12: First I thought I had "count" in the right place for it to be reset.  In this case I copied/patterned after "triggerred"
```
    private class WakeupTimerTask extends TimerTask {
        // Sleep message after max battery awake duration or 20 seconds
        private boolean triggered;
        private int count;
        private final ZWaveWakeUpCommandClass wakeUpCommandClass;
        private int awakeMax = (controller.getSystemMaxAwakePeriod() * 2);
```
```
           count = 0;
            triggered = false;
```
What is going on here: posssible tl;dr
My testing uncovered that if initialization does not complete the first time, the current kickQueue on the subsequent wakes (user pressing the button) is 250 ms before the device is set awake, so nothing happens, ironically until the "Go to sleep" message is queued.  That triggers the one configuration message stuck in the queue, immediately followed by the "go to sleep" message.  I have logs in the pdf document that show this.  

Initially I had this at 0.5 seconds (count=1), but since it is hopefully going to be rare after everyone cranks up the maxAwakePeriod, I moved it 2.0 seconds (count=4), so that it did not potentially disrupt "normal" business at least for me (88% of my activity is under 2 seconds). It is only needed once, and even with the default of 5 seconds will still provide good progress on initialization. On my tests, you can get 10-15 messages per second.  The total number of messages needed for initialization varies (Fibaro and Aeotec tend to require a lot), but 30-45 messages will go a long way.  Plus if I can get the doomed AddNode Stop serial message fixed, the initial configuration time at the default setting will be 5 seconds, not 1 second like it is now.

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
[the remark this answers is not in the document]
cdjackson: This should not be required, and it is not related to the management of the wakeup queue so I really would not like it hidden away in here.

I'm still to be convinced it is needed at all as the controller ought to respond in a timely fashion, but it is not part of the wakeup queue management in any case.

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
mhilbush: I see. Thanks for the explanation. If @cdjackson agrees, to me this would be worthy of a (concise) comment in the code explaining what's going on.

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -119,11 +119,11 @@ public void initialize() {
...
         }
 
-        param = getConfig().get(CONFIGURATION_MAXBATTERYWAKEDURATION);
+        param = getConfig().get(CONFIGURATION_MAXAWAKEPERIOD);
         if (param instanceof BigDecimal) {
-            maxbatteryawake = ((BigDecimal) param).intValue();
+            maxAwakePeriod = ((BigDecimal) param).intValue();
         } else {
-            maxbatteryawake = 5;
+            maxAwakePeriod = 5;
```
cdjackson, line 126: Can you check to see what happens if you remove this completely? I think with the recent changes to core that configuration should be set to the defaults if a default is set in the configuration definition.  If so, it would be best to only define the default in one place (ie the XML) - if not, we should move this default definition to a constant and define it at the top of the file.

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -337,7 +337,7 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
                     // TODO: Do we need to set this immediately
                 } else if (cfg[1].equals("inclusiontimeout") && value instanceof BigDecimal) {
                     reinitialise = true;
-                } else if (cfg[1].equals("maxbatterywakeduration") && value instanceof BigDecimal) {
+                } else if (cfg[1].equals("maxawakeperiod") && value instanceof BigDecimal) {
```
cdjackson, line 340: Are you sure this is the best approach here? Do we really need to reinitialise the system if we change this configuration. I don't think you need to do this at all - just set the value of `maxAwakePeriod` since in the node class it always requests it from the controller each time it's used as far as I can see?

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java
```diff
@@ -107,7 +107,7 @@ public class ZWaveController {
 
     private ZWaveInclusionController inclusionController = null;
     private int defaultWakeupPeriod = 0;
-    private int maxbatterywakeduration = 5;
+    private int maxAwakePeriod = 5;
```
cdjackson, line 110: To avoid confusion, I'd suggest to remove the `= 5` from the end of this - it doesn't do anything.

cdjackson (changes requested):
A few more comments...

One other point - please can you add a short description of this config parameter to the README file.

apella12:
Last comment for the day- Cocktail hour !  Despite signing off exactly as it did before it now wants me
`git rebase HEAD~2 --signoff`
`git push --force-with-lease origin apella-zwave-battery-device-optimize`

I spent all afternoon on this and do not want to do anything that will mess it up until it is reviewed.  Last time I issued commands based on the DCO advice, it totally screwed things up.

Anyway most things seem resolved (to me anyway)

Bob

PS- Whoops !! see it is already daylight hours in NZ.  Will address additional comments tomorrow.  I'll make a mistake or say something I regret if I do it now.

cdjackson:
> Cocktail hour

Enjoy :)

> I spent all afternoon on this and do not want to do anything that will mess it up until it is reviewed. Last time I issued commands based on the DCO advice, it totally screwed things up.

Ignore this - just keep making the changes. I'm happy you've signed off the first commit and that's what counts as ultimately all commits will get squashed so we'll loose the history anyway.  Just keep doing what's easy for you :)

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveController.java
apella12: Done

Again just patterning after the defaultWakeupPeriod

`private int maxAwakePeriod;`

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
apella12: I vaguely recall that before I found that, when I changed the maxAwakePeriod on the controller UI page, it did not propagate or show up in the JSON API query. It only takes the controller offline for about 30 seconds.  I'll try without for a test, but leave in case
```
//                } else if (cfg[1].equals("maxawakeperiod") && value instanceof BigDecimal) {
//                    reinitialise = true;      
```

After a test, it does need to be there.  See the log file copied in re: the AddNode comment.  Node 3 was after I changed the parameter to 10 seconds, but still was only awake for 5 seconds.

However, the file below I restored the reinitialization and it was changed to 10 seconds.  Total time to initialize was 7.8 seconds, or 3.8 seconds after the 4 second delay from the AddNode change "One and Done"
[tensecond default.log](https://github.com/openhab/org.openhab.binding.zwave/files/8611811/tensecond.default.log)

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/AddNodeMessageClass.java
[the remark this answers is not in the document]
apella12: I'll remove these changes, but it makes me sad because unless you agree to raise the default maxAwakePeriod to 10 seconds from 5 seconds there will not even be a chance for "One and Done" initialization of battery devices at the default settings.  It also raises the importance of the delayed queueKick() at 2 seconds
[batteryfeedbackchanges.txt](https://github.com/openhab/org.openhab.binding.zwave/files/8611674/batteryfeedbackchanges.txt)
 to get initialization going again.  What I see in this file is 1) the 4 second delay caused by this doomed message (callback=0) 2) The kickQueue() kicking in at 2 seconds on the next awake 3) the whole process getting out of whack with a message timeout, but eventually getting done.  Basically the current situation, but improved a bit with the kickQueue

Also related to a comment below.  Node 3 was done after I changed the parameter in the UI to 10 seconds.  It looked like it was changed from the API/JSON, but was not.   Otherwise the same observations for a second time.

apella12:
Noon update:  finished testing a couple of things (log files in line).  I think I have all the code related items addressed and tested, but would like to consider an increase in the default maxAwakePeriod so this PR has an actual chance of working at the default settings (with user intervention to change maxAwakePeriod it works fine).  Again see log file with 10 second default.
Need to run errands. 

Still TODO:Readme edits

Bob

src/main/java/org/openhab/binding/zwave/internal/protocol/ZWaveNode.java
apella12: Added a comment in the code

apella12:
This is the only thing left that I'm concerned about

> I'm not sure why - with the check to see if there are still frames queued, then it should be fine?

The way the variable timer works/is working the "Go to sleep" message will be triggered by the _**soonest**_ of **either** the `maxAwakePeriod` being attained **or** (when the message queue is empty **and** the device is "done").  The idea, going back to the #1100 discussion, was to provide a backstop when the device went to sleep with a message in the queue. If I were king, which I am not, I would set it a 15, now that the AddNode problem is still around.  The data in my second post was with the `maxAwakePeriod` at 19, but actual average time awake was only 1.5 seconds.  I have since lowered it to 11 on my production system, but I don't do network heals which take the longest.

Bob

Also figured out the signoff problem.  Needed the email in <>'s

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
cdjackson: > I vaguely recall that before I found that, when I changed the maxAwakePeriod on the controller UI page, it did not propagate or show up in the JSON API query. 

I don't think that can be correct. The data gets saved in the config and provided via the REST API and this has nothing to do with restarting the binding.

> It only takes the controller offline for about 30 seconds.

It also does a reinitialisation on all devices which can slow things down. I agree - it's not the end of the world, but it should be avoided if it's not required. 

> After a test, it does need to be there. 

I don't understand why.

What did you change - did you just comment out the two lines that you showed in your comment above?? If so, that won't work and was not what I suggested. My suggestion was to set the `maxAwakePeriod` -:

```
                } else if (cfg[1].equals("maxawakeperiod") && value instanceof BigDecimal) {
                    maxAwakePeriod = xxx
```

If you just comment these lines out, then the `maxAwakeVariable` will only be initialised when you restart the binding which is a pain.

apella12: So I tried two things, both not working from the UI controller "Thing" page (but for different reasons), but both "work" on the UI controller "Code" page.  It appears a change in the `maxAwakePeriod` on the "Code" page reinitializes the controller.  I can't tell if this would cause all the devices to reinitialize because my test system is just the controller, but I suspect it would.  By "work" on the "Code" page I mean when I scan/discover my battery device the `maxAwakePeriod` is used.

My first attempt was a google try and was good enough to pass the VSC and compile errors, but sprung an OH error (relating to Big Decimal and Integer) when I tried to change the parameter on the UI "Thing" page.  (I did mention my java skills are mostly to copy a similar code snippet which is why I used the `inclusiontimeout` directly above)  Anyway I tried
`maxAwakePeriod = (int) value`

The more interesting attempt was when I left off any second line.  This time it updated on the UI "Thing" and appeared to make it the section of code above because at the Debug level.  I get
`2022-05-05 14:30:38.536 [DEBUG] [zwave.handler.ZWaveControllerHandler] - Controller Configuration update controller_maxawakeperiod to 11`

However on the subsequent scan/discovery of the battery device 6 seconds was used.  This was the value set on the "Code" page.  The order of events in the log are 1) startup 2) change to 9 seconds- Thing Page 3) change to 12 - Thing page 4) change to 6 -Code page 5) change to 11 - Thing page 6) Scan/discover Node 9 - Timeout at 6 seconds. The full Debug log is attached
[zwave.log](https://github.com/openhab/org.openhab.binding.zwave/files/8635617/zwave.log)

One related comment.  The maxAwakePeriod is not intended to be changed often.  I have only changed it once on my production system from 19 seconds to 11 seconds in 3 months and that was because I collected the data above.  Just a thought.

Bob

apella12: Looking this over, one possible thought. The `ZWaveNode` class pulls the `maxAwakePeriod` from the `ZWaveController` class.  
```
    public int getSystemMaxAwakePeriod() {
        return maxAwakePeriod;
```
It is created there from a section with javadoc indicating "Constructor", but the configuration change of `maxAwakePeriod` happens in the `ZWaveControllerHandler` class.  Could this be why a change in `maxAwakePeriod` only gets operational after a reinitialization?  If you think that makes sense and could avoid reinitialization, I could try a change tomorrow. 

I do not believe this code is ever used in the `ZWaveControllerHandler`.
```
    public Integer getMaxAwakePeriod() {
        return maxAwakePeriod;
```
Let me know if you have time.  I realize this PR is not "One and Done" for you-for that I apologize, although I hope it is for battery device folks in the future.
Bob

cdjackson: When I looked at this the other day, I saw that the `ZWaveNode` class pulls this from the controller method `getSystemMaxAwakePeriod` - as you saw above. So I assumed that it doesn't seem to be stored in the node, and therefore as it's pulled from the controller each time the wakeup occurs, all we had to do is the set the max wakeup period in the `handleConfiguration` method, and it would automatically be used.

I still don't really understand why that's not the case if you are doing this.

Now... Looking at your code -:

> Anyway I tried `maxAwakePeriod = (int) value`

I'm not sure this will work, since `value` will not (I think) be an integer, it will be a `BigDecimal` and I think you should see a `ClassCastException`.  You would probably need to do something like `maxAwakePeriod = ((BigDecimal) value).intValue();`

I'm pretty sure though that if you update `maxAwakePeriod` at this point, there should be no need to reinitialise the binding due to the node reading this value each time (as above).

Maybe I'm missing something though?

> Let me know if you have time. I realize this PR is not "One and Done" for you-for that I apologize, although I hope it is for battery device folks in the future.


No probs - we'll get it sorted :)

apella12: > No probs - we'll get it sorted :)

First, I do appreciate your time.  I know you have a real job and being in your time zone (with Germany for OH core and me and others in the USA), you could be busy 24 x 7.

This is what I think I intended
1) For the very first time (new Controller) the `maxAwakePeriod` flow goes `controller_serial` --> `ZwaveBindingConstants` --> `ZWaveControllerHandler` --> `ZWaveController` --> `ZWaveNode`.  This seems to work.

For a reinitialization I think it is `ZWaveControllerHandler` --> `ZWaveController` --> `ZWaveNode`.  This also seems to work.

The item I am worried about is that `getSystemMaxAwakePeriod` is pulled from the ZwaveController class, but the `handleConfiguration` method is in the `ZWaveControllerHandler` class. Based on the logs attached above, I actually think the `maxAwakePeriod` is updated in the `ZWaveControllerHandler` class already. What I'm not sure of is how `maxAwakePeriod` gets to the `ZWaveController` so it can be pulled by the `ZWaveNode`. I am likely wrong, but it could explain why it is not working. 

You'll probably laugh or cringe, but I do recall early on in my adventure trying to pull getMaxAwakePeriod() from the `ZWaveControllerHandler` and skip the `ZWaveController` class but the `ZWaveNode` class doesn't 'talk' to it, I even recall adding it as an import to no avail. My skills at java are very limited, as you are obviously aware, but I keep chipping away at it.

Bob

apella12: I might need a hint to get the `maxAwakePeriod` to change without reinitialization.  No new code to post because I haven't gotten it working.  I do think that in the `ZwaveControllerHandler` class the update is working fine. I currently have these lines
```
             } else if (cfg[1].equals("maxawakeperiod") && value instanceof BigDecimal) {
                    maxAwakePeriod = ((BigDecimal) value).intValue();
                    logger.info("maxAwakePeriod {}", maxAwakePeriod);                  
                }
```

and get `2022-05-06 16:34:39.704 [INFO ] [zwave.handler.ZWaveControllerHandler] - maxAwakePeriod 12`

What hasn't worked yet is getting the `maxAwakePeriod` from the `ZWaveControllerHandler` class vs the `ZWaveController` class in the `ZWaveNode` class.  I get a `NullPointerException` when I try to use the commented out line below.(it is line 1464-so it a match)
```
2022-05-06 16:26:33.353 [ERROR] [alization.ZWaveNodeInitStageAdvancer] - NODE 13: Error in initialization thread
java.lang.NullPointerException: null
	at org.openhab.binding.zwave.internal.protocol.ZWaveNode$WakeupTimerTask.<init>(ZWaveNode.java:1464) ~[bundleFile:?]
```
```
        private int awakeMax = (controller.getSystemMaxAwakePeriod() * 2);
//        private int awakeMaxTest = (controllerHandler.getMaxAwakePeriod() * 2);
```
I imported the `ZWaveControllerHandler` class into the `ZWaveNode` class and also added the line below to get it to compile ok.  I also get the right javadoc description when I hover over the item.
`private ZWaveControllerHandler controllerHandler;`

I tried changing the get from `Integer` to` int` with on the " get" no success since the working one was `int`
```
public int getMaxAwakePeriod() {
        return maxAwakePeriod;
```
    }
Done for the day here. Will think about it and try again tmw.

Bob

apella12: I'm not going to be able to get the `maxAwakePeriod` from the `ZWaveControllerHandler` class into the `ZWaveNode` class when the parameter is changed via the UI Thing page (using the method `getMaxAwakePeriod`).  The method I am using `getSystemMaxAwakePeriod` in the `ZWaveController` class needs the controller to be reinitialized to be updated. 

 I had tried this unsuccessfully before but forgot to delete the method in the `ZWaveControllerHandler` as it is unneeded in this PR.  I do not get errors in VSC or with Maven, just the NPE when I try to call the method.  I'm guessing it has to do with trying to reach a "higher" directory because when I just use the class name, I get a message that the method in the `ZWaveControllerHandler` needs to be static. 

As noted above, this parameter should really not be changed, once it is high enough so initialization and heals are completed in one awake period.

Sorry,
Bob

cdjackson: I had missed that this was picked up in the controller, and not controller handler.

It should be easy enough to add a new method to the controller (eg setConfig) to update the controller configuration without restarting. eg break the constructor and move all the config stuff into a separate public method (eg called `setConfig(Map<String,String> config)`) and then call that from the constructor, and then I think this can also be called from the handler. I assume all the config names are the same so it should just be a case of passing in the updated config parms.

If you're not keen to do this, then no probs, we'll go with it as is...

apella12: I do think it will not be a problem as is (really should not be messed with that frequently) and am not keen at this point.  However, since I'm trying to learn a thing or two let me look at your suggestion in the morning. and I'll let you know if I think I could figure it out. 

Bob

apella12:
Clicked the commit to soon.  I'm using the Github desktop and VSC and it insists on changing .classpath and .project every time I edit in VSC.  I have removed them religiously from all the other commits and dozens of times on my local copy.  Hopefully these can be deleted from the PR without starting over, but let me know.  Otherwise I think all the changes have now been addressed

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
```diff
@@ -69,6 +69,8 @@ public abstract class ZWaveControllerHandler extends BaseBridgeHandler implement
     private Integer secureInclusionMode;
     private Integer healTime;
     private Integer wakeupDefaultPeriod;
+    private Integer maxAwakePeriod;
+    private Integer maxAwakePeriodUpdate;
```
cdjackson, line 73: This doesn't need to be defined here. From what I see, this is only used in one place so it can be a local variable rather than a field in the class.

cdjackson:
Thanks @apella12 I just have one more small comment then we'll get this merged - sorry for all the pain, but thanks for pressing this through :)

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
apella12: Without researching local and field, I moved this to the only place it is used, next to another variable that is only used there. Hopefully that is correct.  If not I'll look at in the AM.

Not a pain, trying to learn and help the OH community at the same time.

Bob

cdjackson:
Sorry - it can really go further down. It actually doesn't need to be a variable at all as that avoids boxing/unboxing the `int` to an `Integer` (since BigDecimal.getValue returns an int, and not an Integer).

See the following code.

https://github.com/cdjackson/org.openhab.binding.zwave/blob/apella-zwave-battery-device-optimize/src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java#L339

src/main/java/org/openhab/binding/zwave/handler/ZWaveControllerHandler.java
cdjackson: Thanks - I created a suggestion in another comment :)

Basically it can go right down into the little bit of code where it's actually used - and given it's only created as a variable and then passed to another method, it actually doesn't need to be created at all. Take a look at the code I referenced.

Thanks.

apella12: Done

apella12:
FYI- I have tentatively put together a duplicate PR to this one that should pass both checks, is based on today's zwave binding (I did manage to rebase my fork) and only has one commit with the six file changes.  Don't know if it will help, hurt or confuse the process, so will await advice.

I'm still trying to figure out the rebase of a PR by practicing on the `Add Node Stop Fine Tune`.  I did get it to work once (after 1762) but it created more commits and the second time I messed up (it included the DB update) and had to start over.  I could also try to rebase this PR first if that is preferred, now that I have a backup plan, but that will add to the commits. I thought I would hold off in case there are other DB updates or you want additional changes to this one.

Bob

cdjackson:
Thanks @apella12 looks good and sorry again for taking so long to get this merged.

apella12:
No problem.  Thanks for your patience with my code.  Hopefully a few less forum posts on battery initialization will save a bit of your time going forward. 
Bob

mhilbush:
@apella12 I finally got around to including a battery powered device into my network. It completed initialization without me having to do a single wake up of the device. This change rocks!!! Thanks guys for getting this into the binding!

PR #1762, opened 2022-05-12
AddNode Message Class Fine Tune

apella12:
The problem addressed with this PR is that the second Add Node stop message (when “complete”) has Callback=0 because the Silabs specification only requires a controller Response, however the serial message (4A) is set-up to require a Node Request.  Node Request is needed on the serial message because the first time the 4A serial message is sent a Node Request is required.  Since the serial message 4A can’t be configured both ways (at least with my level of knowledge), this PR separates the second calling with a timeout reflecting the Response nature in line with the timeout documentation below. The 0.5 second timeout at twice the default is a proxy for the controller response.  There’s going to be bigger problems if the Response doesn’t happen in 0.5 seconds, but a timeout up to 1 second will not cause the battery device initialization problem described below.

* <h2>Timeouts</h2>
 * <p>
 * A timer thread manages timeouts - different times are used for the different stages of a transaction.
 * Defaults for each timer are as follows -:
 * <ul>
 * <li><i>RES</i>ponse - should be received within <b>250ms</b> of the <i>REQ</i>uest</li>
 * <li><i>REQ</i>uest - should be received within <b>2500ms</b> of the <i>REQ</i>uest or the controllers <i>RES</i>ponse
 * </li>
 * <li><i>DATA</i> - should be received within <b>2500ms</b> of the <i>REQ</i>uest.</li>
 * </ul>

The issue caused by the current 5 second timeout on a 4A message is that initialization is delayed about 4 seconds during the period the battery device is awake.  There is no issue with mains powered devices since initialization resumes after 4 seconds.  However, for battery devices with a 5 second awake period (in the current design or set at that level with #1760) reduces the time available for the first time initialization to 1 second; not enough.  Subsequent reawakes of a not fully initialized battery device do not pick up smoothly, as documented elsewhere and user frustration ensues.

### DIFF
--- a/doc/aeon/dsb28_0_0.md
+++ b/doc/aeon/dsb28_0_0.md
@@ -62,6 +62,7 @@ The following table summarises the channels available for the DSB28 -:
 | Electric meter (volts) | meter_voltage | meter_voltage | Energy | Number | 
 | Electric meter (kWh) | meter_kwh | meter_kwh | Energy | Number | 
 | Reset Meter | meter_reset | meter_reset | Energy | Switch | 
+| Reporting Interval- Group 1 | config_decimal | config_decimal |  | Number | 
 | Electric meter (watts) 1 | meter_watts1 | meter_watts | Energy | Number | 
 | Electric meter (amps) 1 | meter_current1 | meter_current | Energy | Number | 
 | Electric meter (volts) 1 | meter_voltage1 | meter_voltage | Energy | Number | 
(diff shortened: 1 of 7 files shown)


apella12:
I extracted from my recent initializations and some forum initializations I had on my local machine in the attached file.  Only the period from the second AddNode Stop to the first TX after the command timeout (generally `Manufacturer_Specific_Get`). It is repetitive but I wanted to show the consistency, the timeout is not a fluke.  Within the file I noted the time for the "Response", the longest being 0.043 seconds.
[AddNode Stop Issue.txt](https://github.com/openhab/org.openhab.binding.zwave/files/8680372/AddNode.Stop.Issue.txt)

Also aligning the `Manufacturer_Specific_Get` in the text file with the Zniffer communication shows nothing is sent to the Node.
[addnodestopzniff.pdf](https://github.com/openhab/org.openhab.binding.zwave/files/8680566/addnodestopzniff.pdf)
Realize this may languish, but I think it is still important regarding workable default maxAwakePeriod defaults.

Bob

apella12:
I need to delete this.  I was using it to try the rebase.  It worked once, but not the second time (pulled in a BD update)

Issue #1775, opened 2022-06-16
Add support for setting default tone in COMMAND_CLASS_SOUND_SWITCH 


kennetn:
For the Aeotech doorbell it would be nice to be able to change the tone played when the doorbell is activated from the remote button (ringer).

Currently the COMMAND_CLASS_SOUND_SWITCH does not support changing default tone with the CONFIGURATION_SET command.
The CONFIGURATION_SET command is used to set both volume and default tone.
The COMMAND_CLASS_SOUND_SWITCH allows setting the volume by setting the default tone as 0 indicating that it should not be changed when the volume is changed. 0 is specified to always be ignored so this is fine.

To change the default tone we can use 255 as volume to prevent it from changing volume, but there is still an issue with this. 
For some strange reason the Z-ware specification says that when you use the value 255 (0xFF):
>This value MUST indicate to restore most recent non-zero volume setting.
>This value MUST be ignored if the current volume is not zero (0x00). This value MAY be used to set the Default Tone Identifier and do not modify the volume setting

So you can modify "Default Tone Identifier" without changing volume if the volume is not 0.
But if the volume is 0 and you change default tone then it will change volume also. (I tested this and the Aeotech Doorbell does works this way)

I would not consider this a big issue because I guess most people will not have the doorbell muted and it would only affect them if they change the tone while muted. Not a very typical scenario I guess.
However I still tried looking for a solution and the specification says this about volume values:
>All other values are reserved and MUST NOT be used by a sending node. Reserved values MUST be ignored by a receiving node.

Because the reserved values should be ignored by a receiving node, I tried going outside specification and sending 254 in the hope that the doorbell would ignore the value just as we would like. However it turns out that the Doorbell does not follow spec here. It will still set volume to 84 :frowning: So this is not an option.

So we can either 

1. Ignore the issues for now
2. Change to remember the last received volume and resend that when the tone need to be changed. But I am a bit unsure how to to that. 

@cdjackson  What is your opinion on this ?

cdjackson:
> Change to remember the last received volume and resend that when the tone need to be changed. But I am a bit unsure how to to that.

This is probably the best, but not 100% reliable. If you store the last known volume in the command class (as a field) then it will be persisted in the XML file and therefore retained "forever" - unless the XML gets lost. If the XML is lost, then the reinitialisation should retrieve the value from the device and store it back in the command class...

Does that make sense? I think this should be easy - and assuming of course I fully grasp the issue :)

PR #1777, opened 2022-06-18
Add support for setting default tone in COMMAND_CLASS_SOUND_SWITCH

kennetn:
Added new channel type "Sound default tone" and extended explanation of "Sound tone play" channel type.
The sound switch command class does not always allow setting the default tone without changing volume therefore the last know volume is stored. Tested with Aeotech doorbell.

Closes https://github.com/openhab/org.openhab.binding.zwave/issues/1775

Signed-off-by: kennetn <kennetn@gmail.com>

### DIFF
--- a/src/main/resources/OH-INF/thing/channels.xml
+++ b/src/main/resources/OH-INF/thing/channels.xml
@@ -1245,6 +1245,12 @@
             The value 0 will stop any tone beeing played. 
             The value 255 will play the default tone</description>
         <category></category>
+        <state>
+            <options>
+                <option value="0">Stop</option>
+                <option value="255">Play default tone</option>
+            </options>
+        </state>
     </channel-type>
 
     <!-- Sound volume Channel -->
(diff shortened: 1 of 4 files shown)


src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -1241,7 +1241,9 @@
     <channel-type id="sound_tone_play">
         <item-type>Number</item-type>
         <label>Sound tone play</label>
-        <description>The Sound tone that will play. The allowed tone numbers depend on the device</description>
+        <description>The Sound tone that will play. The allowed tone numbers depend on the device. 
+            The value 0 will stop any tone beeing played.
```
cdjackson, line 1245: It might be nice to add these two values as options (ie 0 and 255).

cdjackson (commented):
Thanks. Just one small comment. Also, did you run the formatter over these files? If not, please can you do so - it looks like there are some formatting differences (but don't run it on the whole repo - just in case it creates a load of unnecessary changes :) ).

src/main/resources/OH-INF/thing/channels.xml
kennetn: I have updated the pull request with your suggested changes

cdjackson:
Thanks @kennetn

PR #1786, opened 2022-07-15
Database update

cdjackson:


cdjackson:
I've not checked the state of the device, but I went through and approved all devices that had changes requested - maybe the device is still being edited. If it's not approved then it will not be in the export.

robbert0h:
Ok this is unfortunate. I had not made any changes to the device definition, however I did upload another version of the device manual. Apparently that needs review too and causes the whole export to fail if it's not flagged?

Anyway, requested review now. Would it be too much trouble to run another export so that I have something to test tomorrow? Thanks.

cdjackson:
I know this is a pain, but previously we had the opposite, and that also causes trouble…

I’ll run another export now - no probs.

PR #1798, opened 2022-08-22
Adapt to core changes (addon.xml)

J-N-K:
Depends on https://github.com/openhab/openhab-core/pull/3050

Signed-off-by: Jan N. Klug <github@klug.nrw>


cdjackson:
I've not followed this change in core and I can't remember if this file is currently/previously used for anything more than as information to the UI, but can you confirm that this doesn't break anything - ie it doesn't require the latest core does it?

J-N-K:
I believe as long as there is no configuration on binding level it'll work without binding.xml. But addon.xml is only required when the core PR has been merged, which depends on JDK17, so this is an issue for OH4, not the next release. At least binary compatibility will be lost then anyway because we'll upgrade to Karaf 4.4/OSGi R8. If maintaining source compatibility is required, we can leave binding.xml in place, it's just ignore by the new core.

cdjackson:
That's fine - I don't think I'm too worried about backward compatibility in future - just that we don't break existing users of OH3.x stable who might want to use the newer binding for newer devices.

Thanks.

PR #1810, opened 2022-10-04
Corrected entry for Heatit ZM Single Relay

SimBri:
Added "Heatit ZM Single Relay" in the table in things.md with link to existing markdown documentation.
Also corrected a typo and wrong category description.


cdjackson:
Note that the database will overwrite the files relating to the device, so I'll merge this, but it will then be changed back in a few days unless the database is updated.

https://opensmarthouse.org/blog/zwave_device_database_guide

PR #1812, opened 2022-10-10
Reduce logging level

cdjackson:


apella12:
I put this as INFO to give an early notice that the maximum awake period was being reached without going into Debug and since it can be changed (as an advanced option).  But since we have gone over your logging preferences before I,  sadly, understand.

cdjackson:
Understood. I think though that it doesnt fit within the OH guidelines for logging as it is a bit too prolific. Another option might be to make a flag so that it only happens once or something like that?

apella12:
Okay, I'll think about it

PR #1837, opened 2023-01-11
Database update

cdjackson:


apella12:
I have a question.  I see there is a 3.4.1 branch, but none of the recent DB updates are included. Are there any discussions about keeping that current (for a while)?  The heart of the problem is the 4.0 snapshots need Java17 and most the non-developers are on 3.4 -java (including the two in the forum today).

cdjackson:
The 3.4 branch isn't being built, except for the releases. I did mean to do an update for 3.4.1, but there was limited notification about that release - just a statement from Kai that if noone objects he'll do the update "later today" (or something to that effect) and given my timezone I missed that until it was too late. I will try and cherry pick over the latest before 3.4.2 if there is one...

PR #1850, opened 2023-02-24
Add Weather (as Humidity) and low heat alarms

apella12:
Several newer temperature and humidity sensor devices (Zooz zse44, Aeotec ZWA009 and Aeotec ZWA039) have high and low humidity alarms using 0x10 "Weather" and also a low temperature alarm.  This adds PR adds both.  Separately, channels "temperature_cold" and "alarm_humidity" will need to be added to the Zwave database drop down menu.


cdjackson:
Thanks @apella12

PR #1866, opened 2023-05-05
Database update

cdjackson:

### DIFF
--- a/doc/namron/4512715_0_0.md
+++ b/doc/namron/4512715_0_0.md
@@ -1,0 +1,228 @@
+---
+layout: documentation
+title: 4512715 - ZWave
+---
+
+{% include base.html %}
+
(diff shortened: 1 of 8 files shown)


apella12:
It looks like someone has gone through some of the alarms (not me).  The change from alarm_burglar to alarm_tamper are all good. All the devices I have send "3", not "1" or "2", so I assume the changes to devices I do not have should be okay too. 

The changes from alarm_power to alarm_battery should also be positive, however, I made that change a couple of months ago on some zooz devices and they all are still NULL.  (Shouldn't they get set to OK during a restart?) Particular concern is the ZSE44 where a parameter can set the trigger level, the battery went below that and no status change.  I had it on my list to call zooz and ask if the low battery was via notification (0x71) or something else, but was sidetracked.

PR #1898, opened 2023-11-24
Database update

cdjackson:


doc/widom/ums2_0_0.md
```diff
@@ -1,281 +0,0 @@
----
-layout: documentation
-title: UMS2 - ZWave
----
-
-{% include base.html %}
-
-# UMS2 WiDom Smart Roller Shutter
```
lovery, line 8: I have WiDom Smart Roller Shutter device and in openhab 4.0 it was working, but in 4.1 - don't (currently my device is shown to be ONLINE, but when I open the thing details page it shows only the status - online and nothing else, like name, thing type, no tab channels ... 
Screenshot from 2024-04-23 12-17-45
Screenshot from 2024-04-23 12-17-57
). How I can make my device work again?
Also in the [documentation](https://www.openhab.org/addons/bindings/zwave/doc/things.html) of z-wave in the list of supported devices - WiDom [UMS2](https://www.openhab.org/addons/bindings/zwave/thing.html?manufacturer=widom&file=ums2_0_0.html) device is present, but the link leads to a page saying `Cannot render ...`

lovery: @cdjackson I made some research and found that the resource file for widom ums2 device [is moved to mwmc7 file](https://github.com/openhab/org.openhab.binding.zwave/pull/1867/files#diff-07bd984d00f858895d0c3c4944c3bbed7a6b1ddfd1c055c68890342b1cdef381).
Is this an intentional change - removing support for ums2 devices?

If not I can make a PR for returning the support for widom ums2 devices and also adding mwmc7 in the supported things list.

cdjackson: II'm not really sure what the difference in the devices is, but if they are different, then there should be a different entry in the [database](https://opensmarthouse.org/zwavedatabase). Please don't do this through a PR - it won't be accepted as the next database update will overwrite it.

lovery: @cdjackson - the 2 devices are present in the z-wave database : [ums2](https://opensmarthouse.org/zwavedatabase/1004) and [mwmc7](https://opensmarthouse.org/zwavedatabase/1565)
Can I help in some way to fix this? If not PR, then may be issue?

cdjackson: I'm still not really sure what the problem is to be honest. If they are both in the database, then they will both be updated in the binding. If there are changes required, then the changes need to be made directly in the database - then the binding is updated automatically when the database is exported (through PRs just like the one we are commenting on here).

So, please feel free to make any changes directly in the database. Probably you don't have an account, so please create one, then open a ticket there to request edit access and I'll update your account.

PR #1910, opened 2023-12-30
Add SDK 7 and UI Debug comments.

apella12:
Add comment regarding supported controllers and new location for log setting

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ The binding uses a standard Z-Wave serial stick to communicate with the Z-Wave d
 ## Supported Things
 
 The ZWave binding provides support for a large number of devices (currently 802 devices from 114 manufacturers).
-See the [full list of supported things](doc/things.md).
+See the [full list of supported things](doc/things.md). Note: As of OpenHAB 4.1 Controllers based on Silabs SDK 7 are partially supported.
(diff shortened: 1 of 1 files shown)


apella12:
Also, can you review this for accuracy and input; I want to make sure folks know that the LR aspect of the 800 controllers requires special inclusion.

Example: https://community.openhab.org/t/zwave-800-extender-to-utilize-range-benefits/152602
```
The 800 chip “Long Range” capability is misleading.  Long Range Zsticks have dual frequency ranges. One set of frequencies are for “classic” Zwave and have the same range as 700 chip (and 500 chip) controllers. Long Range uses a separate set of frequencies with a spread spectrum technology that allows for higher power. I found this on the Silabs website and the summary below is extracted from my understanding. To achieve the higher power Long Range capability, the device must be included with the following requirements (these could change over time?):
1)	Long Range nodes have 12 bit addresses (vs classic 8 bit) starting at 256.  This would require a major refactoring on the binding.
2)	Long Range inclusion requires Smart Start.  This is not currently coded into the OH Zwave binding.
3)	Long Range inclusion requires S2 security.  The OH Zwave binding currently only supports S0 security.
4)	Long Range is currently only permitted in North America.
5)	Long Range nodes only ‘talk’ to the controller (no mesh) so the binding needs to relay messages to other nodes (No Association Group messages).
6)	Without all the Long Range requirements nodes will automatically be included in classic mode
```

PR #1927, opened 2024-03-31
Update Multisensor parameters

apella12:
Added version 11 multisensor parameters.  Marked as draft since haven't analyzed enough to know if other changes need to be made. Also untested as I have no device with these parameters.


cdjackson:
Thanks - I think its reasonably safe to merge this... At least it can't do any harm...

apella12:
Thanks!

Also note I opened a ticket on Opensmarthouse.  There is an OpenTherm device from the forum that uses some of the added parameters and will need channels.

PR #1940, opened 2024-05-31
Clarify controller support

apella12:
Had some feed back from a forum moderator that this was unclear.  Removed SDK reference and used 700 & 800 chip references, since that is front and center when buying a controller.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ The binding uses a standard Z-Wave serial stick to communicate with the Z-Wave d
 ## Supported Things
 
 The ZWave binding provides support for a large number of devices (currently 802 devices from 114 manufacturers).
-See the [full list of supported things](doc/things.md). Note: As of OpenHAB 4.1 Controllers based on Silabs SDK 7 are partially supported.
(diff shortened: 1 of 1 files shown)


cdjackson:
Is it true that 800 controllers ALWAYS have a fixed firmware, and 700 controllers always have a different fixed firmware?  I've not played with the latest controllers, but certainly in the past it was completely unrelated - you could have a 500 series chip with an old SDK (and in fact that was very normal). I think it's a little different now, but are we sure that this change is really correct and SDK7 = 700, and SDK8 = 800, and it's not possible to have an 800 series chip with SDK7?

apella12:
No. I can't attest to ALWAYS anything, but both 700 and 800 chips use the SDK7 now.  My 700 chip is on SDK 7.18.2 and my 800 is 7.19.1 (but see there is a firmware of 7.21.3 on the zooz site for the 800). I'm not aware of an SDK8, but there is probably one in development.  I also do recall reading somewhere that Silabs was planning to split the SDK development for the 800 chips, but at least for the present both 700 and 800 controllers are on SDK7.  The original READme is technically more accurate (SDK7 is supported), but people focus on the 700 or 800 chip (because that is how they are advertised and not the SDK being used.

PR #1958, opened 2024-11-25
Added System notification events used by a device

apella12:
The Euro Comet (Device 1561) uses system notifications 1-4 that are not covered in alarm_system channel, but are available in the binding.
Signed-off-by: Bob Eckhoff <katmandodo@yahoo.com>  From that device manual; However I just used the current labeling used in the specification, so they could be used by others
Notification 2024-11-25 145929


cdjackson:
Thanks @apella12

PR #1959, opened 2024-11-28
Address multichannel CC with only EP 0

apella12:
I think this will work based on your guidance.  I just made the minimum amount of changes. As I understand it only the GET will be returned and processed to the node advancer, so it will finish?

There does appear to another community post on the same topic(Shelly mini). As an editorial, I see lots of posts on Shelly they seem to be popular, but have issues.


apella12:
Tested locally on a multichannel device and it still works, so that's positive.  I don't have the problem device.

cdjackson:
Thanks @apella12 - if I understand correctly from what I read on the forum, this is now working and results in the XML being produced? If so, I'll merge...

apella12:
Yes it did work for the problem device to produce a OH XML.   OP has subsequently added device 1648. I also config the multichannel CC on that device to remove.  The PR changes also didn't affect a working multichannel device that I tested.  I think it is good to go unless you see something else.

cdjackson:
Thanks @apella12

PR #1969, opened 2025-01-06
Use the Set Serial API (0x0B) to reset nodeId to 8 bits

apella12:
The other PR was not going to work without major refactoring to allow both 8 and 16 bit node IDs.  Since the binding can not support LR nodes anyway (S2 and Smart Start are also required), this PR sends a message to use 8bit. It worked on a zooz 800 with SDK 7.22 originally using 16bit. (I used Simplicity studio to change the zooz 700 controller on 7.21.41.
NodeID 2025-01-04 205530
NodeID page 2 2025-01-04 205642

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/SetNodeIdTypeMessageClass.java
+++ b/src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/SetNodeIdTypeMessageClass.java
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) 2010-2025 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.zwave.internal.protocol.serialmessage;
+
+import org.openhab.binding.zwave.internal.protocol.SerialMessage;
+import org.openhab.binding.zwave.internal.protocol.SerialMessage.SerialMessageClass;
+import org.openhab.binding.zwave.internal.protocol.ZWaveController;
+import org.openhab.binding.zwave.internal.protocol.ZWaveSerialMessageException;
+import org.openhab.binding.zwave.internal.protocol.ZWaveSerialPayload;
+import org.openhab.binding.zwave.internal.protocol.ZWaveTransaction;
+import org.openhab.binding.zwave.internal.protocol.transaction.ZWaveTransactionMessageBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class sets the NodeId Type to 8 bits
+ *
+ * @author Bob Eckhoff - Initial Conteribution
+ */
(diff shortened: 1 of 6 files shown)


apella12:
Don't know if you are busy or waiting on more information.  In case of the latter here is a sample of the debug files;
1) New OOB Zooz 800 with current binding 
[openhab800new.log](https://github.com/user-attachments/files/18390991/openhab800new.log)
Original 800 current binding
2) Binding with PR (includes device exclusion and inclusion after controller initialization)
([openhab800withfix.txt](https://github.com/user-attachments/files/18390995/openhab800withfix.txt))
Fixed 700 with 8 bit
3) Binding with PR on 500 controller (Suspected the CC 0x0b was not supported- it isn't, but doesn't seem to cause problems). I did add the OR based on the device response however)
[openhab500withfix.txt](https://github.com/user-attachments/files/18391011/openhab500withfix.txt)

cdjackson:
Sorry - I missed that you'd removed the draft tag. I'll try and take a look later today.

cdjackson:
Thanks.

Two points. Firstly, does this work on old controllers or will it (possibly) cause a problem? I'm just wondering if we shouldn't make this conditional and only send it on 800 controllers?

Second point - can you pull out the constants and put them at the top of the file as `static int` rather than using them directly. I appreciate it's a short file, but it's generally better to keep constant defs at the top.

cdjackson:
Thanks.

Yes, those are the constants I meant. Instead of calling them `payload0` etc I'd give them a meaningful name - so maybe `CMD_SETNODEID` for pl0 and `NODEID_8BIT` for the pl1?

src/main/java/org/openhab/binding/zwave/internal/protocol/serialmessage/SetNodeIdTypeMessageClass.java
```diff
@@ -30,12 +30,15 @@
 public class SetNodeIdTypeMessageClass extends ZWaveCommandProcessor {
     private final Logger logger = LoggerFactory.getLogger(SetNodeIdTypeMessageClass.class);
 
+    private static byte cmdSetNodeIdType = (byte) 0x80; // Subclass of SetUpZwaveAPI for NodeID Type
+    private static byte nodeIdType8Bit = (byte) 0x01;  // 1=8 bit, 2=16 bit NodeID Type
```
cdjackson, line 34: Sorry to be a pain, but please make these `final`. Also, as per my suggestion, please use the convention of upper case for constants, delimited with underscores rather than camel case.

cdjackson: Thanks, and no worries - I appreciate the changes and sorry to be picky :)

apella12:
1) I thought this was ready to go. Maybe you are just busy. Maybe there is something else.
2) When/if you have time would it be possible to cherry pick the recent ZW DB updates (Since OH5) and this PR back to an OH4.3 version, so it could get picked up if/when a OH4.3.3 is released? OH5.0 requires Java 21 and 64bit OS that might be too much for the some that stick to a stable release.  There are workarounds I've written in the forum, but require some java skill or the Silabs PC controller, if this is too much work

cdjackson:
Sorry @apella12 - I was sure I'd merged this (too much going on).

I did try and cherry pick the DB before the last 4.3.2 release, and it caused some problems that I didn't have time to sort out. I'll have another look before the next release.

PR #1983, opened 2025-04-24
Add semantic equipment tags dynamically

andrewfg:
#### Background

Originally I had used an auto tagger program to scan the thousand plus thing-type xml files and automatically add a static semantic equipment tag to them. However @cdjackson pointed out that these xml files are generated dynamically, so such an approach would not stick. So I abandoned that approach..

#### New Approach

Instead of (trying to) apply tags at design time, the new approach applies them dynamically at run-time. It parses the node Thing's SPECIFIC and GENERIC properties and applies a semantic equipment tag based on those properties..


cdjackson:
Unfortunately I don't see any point in merging this. The database export will simply overwrite this next time it's done, so either we need. to find a way to run this within the export scripts, or as a separate PR after each export (which is probably not a good idea).

Probably the best approach would be to find a way within the database to add this support, so that users can change them rather than use a rule based system (ie as I assume you've done in the script). That's not going to be easy though as the database uses a commercial web app framework that I don't subscribe to now, so any significant modification is far from easy.

I'm open to suggestions on how to do this...

andrewfg:
Ah. I was not aware that this is a dynamic operation. Let me think about it..

andrewfg:
See https://github.com/openhab/openhab-addons/pull/18694#issuecomment-2888998233

andrewfg:
@cdjackson I completely changed the approach for applying the semantic equipment tag. So hopefully this is now a workable solution.

andrewfg:
@cdjackson any thoughts on this -- now revised -- PR?

cdjackson:
Sorry - I've been travelling a lot between New Zealand and Europe recently :(

This looks fine to me - thanks.

PR #1990, opened 2025-08-03
Add GitHub Actions workflows

wborn:
Adds GitHub Actions workflows similar to what is used in the other openHAB Java repositories.

---

When this is merged, can you also add the GHA build to the required status checks in the branch protection rules and remove the Jenkins build from it?
After that we can disable the PR build in Jenkins like was done in the other repositories.


kaikreuzer:
Thanks @wborn, all done!

PR #1991, opened 2025-08-08
Update parent to 5.1.0-SNAPSHOT

wborn:
It will cause issues eventually when the wrong parent is used.


apella12:
fyi- My VSC development environment needed this to avoid thousands of errors. It would be good to get this merged

PR #1996, opened 2025-09-02
Skip static init and NIF request for restored nodes

apella12:
When a node is restored from a config file, initialization now jumps directly to the DONE stage and skips the REQUEST_NIF step to reduce network congestion. Persistence is relied upon to restore values, assuming the node on the controller matches the restored node.
Probably a little more controversial... I have been using both ZUI and this binding for a couple of years now. One thing I noticed right away is that it only does a ping on startup for devices already in their "Store". The zwave-js has applied (or at least targeting) ZWA certification, so this must be allowed.  This PR allows devices that are recovered from XML files to basically go to "Done". In one case I had had one device without an XML (The battery devices in my test system have no XML, so they go to wait NIF). The only devices get a ping and go to complete.  The one device without an XML goes through all the steps.
[WSL Ping only Test with OH5-main network.txt](https://github.com/user-attachments/files/22104981/WSL.Ping.only.Test.with.OH5-main.network.txt)
In this test on another controller only pings of listening devices.
[WSL 800 Basement Test  Ping only.txt](https://github.com/user-attachments/files/22104985/WSL.800.Basement.Test.Ping.only.txt)
About 22 nodes in 2 seconds.  As OH in general gets bigger this would help IMO.  Also since the devices are DONE, they could be reinitialized or Healed outside the startup window, if something doesn't look okay.

### DIFF
--- a/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
+++ b/src/main/java/org/openhab/binding/zwave/handler/ZWaveThingHandler.java
@@ -467,8 +467,10 @@ public class ZWaveThingHandler extends ConfigStatusThingHandler implements ZWave
         }
     }
 
+    // Start polling with a random initial delay, but right after the thing is initialised (DONE).
+    // 30 seconds mimimum, 90 seconds maximum
     private void startPolling() {
-        startPolling(pollingPeriod * (int) (1000 * Math.random()));
+        startPolling(30000 + 60 * (int) (1000 * Math.random()));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


cdjackson:
It's not just about automation - it's about state in general. The UI will also be incorrect. Yes, you have persistence, but that restores the state at some previous time - if OH has been down, then it's likely that (some!) states will be different. Does this matter - who knows - it depends on your situation, but the user will not know about those changes until the association fires on the next change - or if they have some slow poll running which is probably an hour or potentially much longer away.

One example, if the power went down, then the state of lights / sockets etc will have likely changed - persistence will show the socket still on (from before the power cut), and the power cut may have turned it off. You need to poll the device to update this state - when the poll is complete, the item state changes (since persistence will have restored the state to on, and the poll will change that to off). I have exactly this issue - I want to know when the state changes so I can do something about it and the only way to do this is to poll the device (I will add that in my case it's the zigbee binding, but the situation is the same).

> If it would make a difference I could look at adding a new poll of linked channels for powered nodes

I think this would be the best approach - thanks.

Issue #1909, opened 2023-12-26
Shall we run mvn spotless:apply ??


andrewfg:
While working on a PR for this repository, I noticed that running 'mvn spotless:apply' results in many tens of files being reformatted.

As it is now common practice in openhab-core and openhab-addons to keep all source files spotless compliant, I wonder if we should do the same here? => If there are no objections, I would myself be quite happy to create a PR with spotless applied.

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

cdjackson:
This has been discussed in the past and I'm personally not keen as it looses a lot of history and IIRC from previous discussions, also changes the XML files which will be changed back with the next update.

andrewfg:
> changes the XML files which will be changed back with the next update.

From memory, it seemed to me that most of the format corrections were applied to java files and not xml files. Or??

cdjackson:
Ok, maybe I remembered that incorrectly, or possibly an exception was added to avoid changes. In any case the main point was I’d prefer not to loose the history and just do the updates as the files are touched rather than in one large reformat.

Nadahar:
It's quite painful as it is, because when you make a change to a file, and run spotless after editing (as I always do, because the spotless formatting isn't something that makes sense to be as a human, so I can't produce it manually), I end up having to revert changes to lots and lots of files that I didn't touch. A huge number of these files just have one or two indentation changes, so applying that to those files at least, wouldn't really hurt history much, and would reduce the number of files that must be reverted substantially.

PR #2000, opened 2025-12-11
Apply trivial spotless changes

Nadahar:
I just read #1909 after having been frustrated by having to revert a lot of changes after applying spotless to my own changes. This doesn't completely take care of #1909, but is very close to do so. What I did is to run spotless on the code, and then revert the changes that weren't what I considered "trivial". The rest is in this PR, I don't believe this commit will make history any more difficult to track, except perhaps to a very slight extent in `pom.xml` since some sections have been shuffled around. It that's considered a problem, I can revert the changes to `pom.xml` as well.

Merging this would make the repo much easier to work with when using spotless to fix your own changes., because changes to unmodified files will be minimal.

The vast majority of these changes are indentation changes, with a few whitespace corrections. The POM changes are merely about the order of elements.


kaikreuzer (approved):
I think these changes make sense and help contributors massively.
It should not have any significant impact on the git history of files - changes can still nicely be traced.

